PR #24, opened 2023-02-14
fix: addpkg message validation

luciorubeens:
When trying to sign a message with type `/vm.m_addpkg` it throws the error `UNSUPPORTED_TYPE` because it's missing a `break` in the validation.


jinoosss (approved):

dongwon8247:
Thank you!

PR #35, opened 2023-04-01
Fix url parse parameters

sienrik15:
### 1.- Show URL parsing error location.search value 
err_3

### 2.- Show solution
solution

Other example:
image


jinoosss (approved):
Thanks for the suggestion

PR #36, opened 2023-04-13
feat: [ADN-1] Changing the Account and Wallet Signer Structure

jinoosss:
I changed the structure of the Adena module.

Changed the concept of signers (responsible for signing functions.)
- Create multiple accounts with one keyring
- Manage multiple keyrings in the wallet


.gitignore
```diff
@@ -45,4 +45,5 @@ node_modules/
 
 result.info
 deploy/*
-deploy-latest/*
\ No newline at end of file
+deploy-latest/*
+packages/*/types
```
luvhaeun, line 49: @jinoosss Why is this in here?

packages/adena-extension/src/components/background/background.tsx
```diff
@@ -10,143 +10,143 @@ import { useWalletBalances } from "@hooks/use-wallet-balances";
 import { useLoadAccounts } from "@hooks/use-load-accounts";
 
 interface Props {
```
luvhaeun, line 12:
```
type BackgroundProps = React.PropsWithChildren<Record<string, never>>;
```

packages/adena-extension/src/components/background/background.tsx
```diff
@@ -10,143 +10,143 @@ import { useWalletBalances } from "@hooks/use-wallet-balances";
...
-    children?: React.ReactNode;
+  children?: React.ReactNode;
 }
 
 export const Background = ({ children }: Props) => {
```
luvhaeun, line 16:
```
export const Background: React.FC<BackgroundProps> = ({ children }) => {
```

packages/adena-extension/src/hooks/use-add-account.tsx
```diff
@@ -1,6 +1,7 @@
 import { useRecoilState, useResetRecoilState } from 'recoil';
 import { WalletState } from '@states/index';
 import { useAdenaContext } from './use-context';
+import { SeedAccount } from 'adena-module/src';
```
luvhaeun, line 4: @jinoosss Is src really necessary?

.gitignore
jinoosss: I think this should be removed.
I'll modify it so that the type is generated in the build path.

packages/adena-extension/src/hooks/use-add-account.tsx
jinoosss: I think I need to fix this.
I think we need to fix the type build part of the adena-module.

luvhaeun (approved):

PR #38, opened 2023-04-14
feat: [ADN-18] Create a token balance component

jinoosss:
I've written a few components to display the token balance.

[Components]
- common/HighlightNumber: Component to highlight text in integer parts
- common/TokenBalance: Combination of token denom and token number
- main/MainTokenBalance: Token balance on the main screen
- main/MainTokenListItemBalance: Token balance in the token list on the main screen


luvhaeun (approved):
LGTM

PR #40, opened 2023-04-18
feat: [ADN-19] Create a token list component

jinoosss:
Create a component that shows a list of tokens.
It was created as common because it is used on the main screen and the token transfer page.

[Components]
- common/TokenList
- common/TokenListItem
- common/TokenListItemBalance (moved from main/MainTokenListItemBalance)


jinoosss:
Thanks for the review.
I added a default viewport to the storybook. (extension size: 360 x 540)
And I modified the styling of the token list items.

luvhaeun (approved):

PR #42, opened 2023-04-18
feat: [ADN-35] Create the Manage Token page token list component

jinoosss:
Create the Manage Token page token list component.

[Component]
- common/toggle
- manage-token/manage-token-list
- manage-token/manage-token-list-item
- manage-token/manage-token-list-item-balance

### DIFF
--- a/packages/adena-extension/src/components/manage-token/manage-token-list-item/manage-token-list-item.tsx
+++ b/packages/adena-extension/src/components/manage-token/manage-token-list-item/manage-token-list-item.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { ManageTokenListItemWrapper } from './manage-token-list-item.styles';
+import { ManageToken } from '@components/manage-token/manage-token-list/manage-token-list';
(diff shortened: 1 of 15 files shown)


packages/adena-extension/src/components/manage-token/manage-token-list-item/manage-token-list-item.tsx
```diff
@@ -0,0 +1,47 @@
...
+    main,
+    tokenId,
+    logo,
+    name,
+    balanceAmount,
+    activated
+  } = token;
+
+  return (
+    <ManageTokenListItemWrapper onClick={() => onToggleActiveItem(tokenId)}>
```
luvhaeun, line 23: Is the behavior the same when the item is clicked as when the toggle is clicked?

jinoosss: Only the toggle button event exists.
The click event has been removed.

jinoosss:
Thanks for the review.

luvhaeun (approved):

PR #50, opened 2023-04-21
feat: [ADN-24] Create the transaction history component

jinoosss:
This is the component used on the transaction history page.


luvhaeun (approved):
LGTM

PR #51, opened 2023-04-23
feat: [ADN-3] Change state variables in wallets and accounts

jinoosss:
- Account lists are no longer stored separately in Storage due to wallet structure changes.
- Changed the key that identifies the current account. Address -> AccountID
- Changed the way we read the address of an account. Previously, it was stored as the address itself, but now it is generated with the public key and prefix.
- Changed the token and balance data management structure to implement the balance inquiry function by token.
- Divided Provider (Service function management, initial data management)
- Removed unnecessary code.


packages/adena-extension/public/manifest.json
```diff
@@ -34,6 +34,6 @@
     }
   ],
   "content_security_policy": {
-    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; default-src 'self'; img-src 'self' data:; font-src https://fonts.gstatic.com; style-src 'self' 'unsafe-inline'; style-src-elem 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net; connect-src 'self' data: https://* http://localhost:* http://127.0.0.1:*; prefetch-src https://*.openlogin.com; frame-src https://*;"
+    "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; default-src 'self'; img-src 'self' https: data:; font-src https://fonts.gstatic.com; style-src 'self' 'unsafe-inline'; style-src-elem 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net; connect-src 'self' data: https://* http://localhost:* http://127.0.0.1:*; prefetch-src https://*.openlogin.com; frame-src https://*;"
```
luvhaeun, line 37: @jinoosss 

Please tell me why this PR includes this fix.

packages/adena-extension/src/common/provider/adena/adena-provider.tsx
```diff
@@ -27,15 +21,13 @@ interface AdenaContextProps {
...
-
-  const axiosInstance = axios.create();
-
-  const [, setCurrentAccount] = useRecoilState(WalletState.currentAccount);
+interface AdenaProviderProps {
+  children: React.ReactNode;
+}
 
-  const [tokenConfigs, setTokenConfigs] = useRecoilState(WalletState.tokenConfig);
+export const AdenaProvider: React.FC<AdenaProviderProps> = ({ children }) => {
```
luvhaeun, line 28: @jinoosss 

```
<React.PropsWithChildren<unknown>>
```

luvhaeun (commented):
@jinoosss 

After fixing the simple fixes I've left, you can merge right away and we can partially explore later.

packages/adena-extension/public/manifest.json
jinoosss: extension to load images by image path.
Previously, image data was generated, stored, and processed with axios, but since the number of corresponding tokens is variable, variable image processing is required (using image paths).

luvhaeun (approved):

PR #58, opened 2023-04-26
feat: [ADN-46] Change the Transaction History API call parameters

jinoosss:
Removed fetching transaction history from gno-client.
Fixed to load Adena API from respository.
Set the parameters to 'from' and 'size'.


luvhaeun (approved):
There seem to be a number of code fixes that are not related to commit messages.

PR #60, opened 2023-04-26
fix: [ADN-57] Fix account addition errors

jinoosss:
Resolve errors that occur when modifying wallet account data.


luvhaeun (approved):
OK

PR #157, opened 2023-05-11
feat: [ADN-37] Add an event listener in Account Change

jinoosss:
Added an event listener that fires when an account is changed.


2woongjae (approved):
I made a minor type correction to your code.
After you check, please merge if there are no special issues.

PR #158, opened 2023-05-12
feat: [ADN-203] Add data migration

jinoosss:
Add version to data stored per storage. (Default: 1)

Adds the ability to proceed with migration by version (version 1 → version 2 → version 3, etc.)

Create backup data for data restoration in case of migration failure.

### DIFF
--- a/packages/adena-extension/src/migrates/migrator.ts
+++ b/packages/adena-extension/src/migrates/migrator.ts
@@ -1,0 +1,12 @@
+import { StorageModel } from '@common/storage';
+
+export interface Migration<R = any> {
+  version: number;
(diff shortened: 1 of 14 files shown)


packages/adena-extension/src/migrates/migrator.ts
```diff
@@ -0,0 +1,81 @@
...
+    private _migrations: Migration[],
+    private storage: Storage,
+    private storageKey: string,
+  ) {}
+
+  protected async serialize(data: StorageModel<Model>) {
+    return JSON.stringify(data);
+  }
+
+  protected async deserialize(data: any) {
```
2woongjae, line 30: `data: string`

2woongjae (approved):
LGTM

PR #190, opened 2023-06-02
Release Adena version '1.7.0'

jinoosss:
- Add an event listener in Account Change.
- Change the address list on send token page
- Add an event listener in Network Change (injection) 
- Make the main page typable
- Change to package version 1.7.0
- Fix loading screen for transaction approvals
- Change the logic of Adding and Editing Address
- Integrate gno-js-client library.
- Fix the logo of tx list in GNOT details page
- Fix logo images not being showed
- Adjust font size of token names in search
- Adjust max width of the path section
- Fix an error of GNOT token page history
- Add data migration
- Fix approval injection popup for ledger account
- Change the unit of storage for address book information
- Change the route of sourcing static token data from gno-token-resource
- Fix no account found error for injection when there is only 1 account in wallet
- Remove the history related to GRC-20 on GNOT token page
- Add token details menu for the GRC20 Token Details page
- Fix font weight bug with error messages while sending tokens


dongwon8247 (approved):
LGTM!

PR #191, opened 2023-06-08
feat: [ADN-207] Implement input data retention of send page

jinoosss:


2woongjae (commented):
@jinoosss I think you are missing the part that clears the history, please check.

PR #192, opened 2023-06-08
feat: [ADN-214] Restrict sending grc-20 to the current address

jinoosss:


packages/adena-extension/src/containers/transfer-input-container/transfer-input-container.tsx
```diff
@@ -50,7 +50,8 @@ const TransferInputContainer: React.FC = () => {
     if (!isNext()) {
       return;
     }
-    const validAddress = addressBookInput.validateAddressBookInput();
+    const validAddress = addressBookInput.validateAddressBookInput() &&
+      (isNativeTokenModel(tokenMetainfo) || addressBookInput.validateEqualAddress());
```
2woongjae, line 54: `(!isNativeTokenModel(tokenMetainfo) && addressBookInput.validateEqualAddress())`
I think this is correct, or is it not? @jinoosss

jinoosss: `(!isNativeTokenModel(tokenMetainfo) && addressBookInput.validateEqualAddress()) `
conditional works fine for non-native tokens, but is always false for native tokens.

PR #195, opened 2023-06-14
feat: Validate Injection message format

jinoosss:


r3v4s (approved):
error code 1002 confirm

PR #220, opened 2023-07-17
feat: Add a custom network feature

jinoosss:


packages/adena-extension/src/repositories/common/chain.ts
```diff
@@ -5,6 +5,66 @@ import { NetworkMetainfo } from '@states/network';
...
+  {
+    id: 'test3',
+    main: true,
+    chainId: 'test3',
+    chainName: 'GNO.LAND',
+    networkId: 'test3',
+    networkName: 'Testnet 3',
+    addressPrefix: 'g',
+    rpcUrl: 'https://rpc.test3.gno.land',
+    gnoUrl: 'https://rpc.test3.gno.land',
```
r3v4s, line 18: shouldn't it be `https://test3.gno.land` ?

r3v4s (approved):
LGTM, check comment about test3 url

packages/adena-extension/src/repositories/common/chain.ts
jinoosss: I'll fix it.

PR #221, opened 2023-07-17
fix: Fix freeze on addpkg execution of DoContract command

jinoosss:


r3v4s (approved):
code looks good, but please test with non-ascii string (such as emoji with utf)

PR #222, opened 2023-07-17
feat: Change the source route for apps and networks

jinoosss:


packages/adena-extension/src/resources/chains/chains.json
```diff
@@ -0,0 +1,59 @@
+[
+  {
+    "id": "test3",
+    "main": true,
```
r3v4s, line 4: key `main` is little bit confused with blockchain terms `mainnet`, `testnet`

r3v4s (approved):
one small comment

packages/adena-extension/src/resources/chains/chains.json
jinoosss: How about 'default'?

r3v4s: works for me

PR #236, opened 2023-07-20
feat: Modify DoContract function response content

jinoosss:


packages/adena-extension/src/common/errors/common/tm2-error.ts
```diff
@@ -0,0 +1,133 @@
...
+  },
+  GAS_OVERFLOW_ERROR: {
+    status: 400,
+    type: '/std.GasOverflowError',
+    message: 'gas overflow error',
+  },
+  UNKNOW_ERROR: {
+    status: 400,
+    type: 'UNKNOWN_ERROR',
+    message: 'Unknown error',
```
r3v4s, line 103: perhaps 'Unknown' -> 'unknown' will be more integrate-ish

r3v4s (approved):
one single comment

PR #237, opened 2023-07-20
feat: Modify the transaction details page

jinoosss:


packages/adena-extension/src/common/utils/client-utils.ts
```diff
@@ -9,6 +9,11 @@ import BigNumber from 'bignumber.js';
 import dayjs from 'dayjs';
 import fetchAdapter from '@vespaiach/axios-fetch-adapter';
 
+export function formatHash(hash: string, num?: number): string {
```
r3v4s, line 12: looks like `num` isn't actually used.

do we need to keep this optional argument?

r3v4s (approved):

packages/adena-extension/src/common/utils/client-utils.ts
jinoosss: Currently we don't use it.
However, we would like to match the form with other formatting functions. 
Does it look good to remove them ?

r3v4s: nah let's keep it for now

we can get rid of it when it becomes unnecessary

PR #238, opened 2023-07-20
feat: Modify the transaction details page

jinoosss:


r3v4s (approved):

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-mapper.ts
```diff
@@ -176,7 +176,7 @@ export class TransactionHistoryMapper {
       valueType,
       date: dateToLocal(date).value,
       networkFee: {
-        value: `${transfer.amount || '0'}`,
+        value: `${fee.amount || '0'}`,
         denom: `${fee.denom}`,
```
r3v4s, line 180: we don't need default value for denom?

jinoosss: We judge an empty string as an error. 
If a denom is required, it will come in the response.

PR #251, opened 2023-08-30
fix: Add territory RPC to the default network list

jinoosss:


r3v4s (commented):
seems like token related typescript files are from #250 
it will be fine if you force push and rebase it.

However, where is `teritori` rpc address? Does it exist outside of repo?

jinoosss:
> seems like token related typescript files are from #250 it will be fine if you force push and rebase it.
> 
> However, where is `teritori` rpc address? Does it exist outside of repo?

I'm managing it statically in this file. (https://github.com/onbloc/adena-wallet/blob/develop/packages/adena-extension/src/resources/chains/chains.json)

r3v4s:
in that case, shouldn't this pr includes change from `false` to `true`
https://github.com/onbloc/adena-wallet/blob/1b55b3d1e80cb063354e81cc63fbd5c5973525b8/packages/adena-extension/src/resources/chains/chains.json#L32

jinoosss:
teritori network is not treated as the default.
The `default` option attempts to treat network information as static and unmodifiable.
teritori network can be modified.

r3v4s (approved):

PR #275, opened 2023-09-06
[ADN-318] fix: Combine the settings menu with the side bar menu

jinoosss:


packages/adena-extension/src/pages/certify/reveal-private-phrase/index.tsx
```diff
@@ -12,17 +12,11 @@ export const RevealPrivatePhrase = () => {
...
-    init();
-  }, []);
-
-  const init = async () => {
+  const seeds = useMemo(() => {
     const mnemonic = wallet?.mnemonic || '';
-    const seeds = `${mnemonic}`.split(' ');
-    setSeeds(seeds);
-  };
+    return `${mnemonic}`.split(' ');
```
harryoh, line 18: I think it's same to `return mnemonic.split(' ');`

harryoh (approved):
👍

packages/adena-extension/src/pages/certify/reveal-private-phrase/index.tsx
jinoosss: I agree. Thanks for the feedback.

PR #281, opened 2023-10-19
feat: add `/vm.m_run` message handler

luciorubeens:
Add handlers for the new message type: https://github.com/gnolang/gno/pull/1001

### DIFF
--- a/packages/adena-module/src/transaction/message-info/message-info-vm.ts
+++ b/packages/adena-module/src/transaction/message-info/message-info-vm.ts
@@ -95,7 +95,7 @@ export const encodeVmRun = (writer: protobuf.Writer, messageInfo: VmRun) => {
     writer.uint32(18).string(messageInfo.send);
   }
   if (messageInfo.package !== undefined) {
-    encodePackage(writer.uint32(18).fork(), messageInfo.package).ldelim();
+    encodePackage(writer.uint32(26).fork(), messageInfo.package).ldelim();
   }
   return writer;
 };
--- a/packages/adena-module/src/transaction/message-info/message-info-vm.ts
+++ b/packages/adena-module/src/transaction/message-info/message-info-vm.ts
@@ -26,6 +26,12 @@ export interface VmAddPackage extends InfoType {
   deposit: string;
 }
 
+export interface VmRun extends InfoType {
+  caller: string;
+  send: string;
+  package?: Package;
+}
+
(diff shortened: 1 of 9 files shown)


packages/adena-module/src/transaction/message-info/message-info-vm.ts
```diff
@@ -80,3 +86,16 @@ export const encodeVmAddPackage = (writer: protobuf.Writer, messageInfo: VmAddPa
   }
   return writer;
 };
+
+export const encodeVmRun = (writer: protobuf.Writer, messageInfo: VmAddPackage) => {
```
r3v4s, line 90: shouldn't it be `messageInfo: VmRun` ??

luciorubeens:
I tried to broadcast a transaction, but I got this error, can you help to investigate?

CleanShot 2023-10-19 at 17 12 51@2x

packages/adena-module/src/transaction/message-info/message-info-vm.ts
```diff
@@ -80,3 +86,16 @@ export const encodeVmAddPackage = (writer: protobuf.Writer, messageInfo: VmAddPa
...
+
+export const encodeVmRun = (writer: protobuf.Writer, messageInfo: VmRun) => {
+  if (messageInfo.caller !== '') {
+    writer.uint32(10).string(messageInfo.caller);
+  }
+  if (messageInfo.send !== '') {
+    writer.uint32(18).string(messageInfo.send);
+  }
+  if (messageInfo.package !== undefined) {
+    encodePackage(writer.uint32(18).fork(), messageInfo.package).ldelim();
```
jinoosss, line 98: This simply seems to be the problem.
Change the size of the buffer in the 3rd argument.
`18` -> `26`

r3v4s: @luciorubeens seems like protobuf has certain buf size value for nth args.
https://www.tabnine.com/code/javascript/functions/protobufjs/Writer/string

luciorubeens: Perfect, thank you guys, now it works

CleanShot 2023-10-20 at 10 00 54@2x

jinoosss (approved):
@luciorubeens 
Thank you for your work.

r3v4s (approved):
looking good

PR #282, opened 2023-10-20
feat: return encoded transaction in Sign Amino response

luciorubeens:
In order to broadcast the transaction after signed, we need the transaction bytes.


jinoosss:
Hi, @luciorubeens .
I checked out what you suggested.

If you want the behavior of gnokey's makeTx, how about adding an adena function?
For example, adena.MakeTx

Do you need a Sign Document or Signature data?

Let me know what you think.

jinoosss:
Hi @luciorubeens ,
I would appreciate it if you could tell me what you think about #283 .

PR #316, opened 2023-12-15
Refactor container & layout

skqksh:
move containers and layout to related page or components folder
Classify components as atoms, molecules, or pages.

remained refactor items
- Separating container logins to hook
- Move container code to index file


jinoosss:
I think it looks good.

But it seems like a good idea to merge it once the structure is clear.
There are still components that have not been separated under the components folder.

skqksh:
> I think it looks good.
> 
> But it seems like a good idea to merge it once the structure is clear. There are still components that have not been separated under the components folder.

ok

PR #317, opened 2023-12-15
chore: Change to work on GitHub action tag creation

jinoosss:


.github/workflows/build-deploy.yml
```diff
@@ -2,7 +2,8 @@ name: Build And Deploy
 
 on:
   push:
-    branches: [release/*]
+    tags:
+      - v*
```
skqksh, line 6:
```suggestion
      - v[0-9]+.[0-9]+.[0-9]+
````

skqksh (approved):

.github/workflows/build-deploy.yml
jinoosss: I think your suggestion is good.

We can use a variety of tags.
The tags we use are usually of the form `vx.x.x`, but as we have seen before, tags like `vx.x.x-rc.x` can be used.

`v[0-9]+.[0-9]+.[0-9].*`

r3v4s (approved):

PR #321, opened 2023-12-21
Refactor css, components

skqksh:


packages/adena-extension/src/pages/certify/export-private-key/ApproachPrivateKey.tsx
jinoosss, line 1: Isn't this a file in component/pages ?

jinoosss (commented):
It looks good to me.
However, it seems better to use values from the theme injected from the provider rather than static data. 
What do you think ?

https://github.com/onbloc/adena-wallet/blob/main/packages/adena-extension/src/app.tsx#L20

packages/adena-extension/src/components/atoms/button/index.tsx
```diff
@@ -1,78 +1,72 @@
...
-  Ghost = 'Ghost',
-  Dark = 'Dark',
-  Danger = 'Danger',
-  Custom = 'Custom',
-}
+type ButtonHierarchy = 'normal' | 'primary' | 'ghost' | 'dark' | 'danger' | 'custom';
 
 export const modeVariants = {
   normal: css`
-    background: ${({ theme }): string => theme.color.neutral[6]};
```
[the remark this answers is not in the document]
skqksh: sure

packages/adena-extension/src/pages/certify/export-private-key/ApproachPrivateKey.tsx
skqksh: I don't think so. 
that's because, 'ApproachPrivateKey.tsx' is not a component that is used repeatedly, but a part of the page.

jinoosss: I understand.
I think it's just a difference of opinion.
I think `component -> page` makes sense to separate UI and data even if it's not reused, for example when creating storybooks and jest files.

jinoosss (approved):
Looks good to me. 👍

jinoosss:
The UI of the screen showing the transaction history seems to be broken.
Is this something we should work on further ?

skqksh:
> The UI of the screen showing the transaction history seems to be broken. Is this something we should work on further ?

Thank you for checking the screen. 
This problem is caused by a style conflict due to classname-based CSS inherited from the parent component.
To prevent similar issues,I will continue to refactor classname into styled-component.

jinoosss (approved):
👍

PR #329, opened 2024-01-21
feat. Update ledger connection UI, fix connection error

skqksh:


jinoosss (approved):
The password setting and all-set pages can be used in common.

What do you think about common separation?

PR #353, opened 2024-01-29
feat. Update web UI

skqksh:


packages/adena-extension/src/components/pages/web/main-header/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
```
jinoosss, line 1: `React` must be imported for the test to run.

jinoosss (commented):
I think this needs some additional UI fixes.
- ~The description of Sensitive Information should be changed on the Add account page.~
- The font size of the title of Sensitive Information should be changed.

jinoosss:
> The description of Sensitive Information should be changed on the Add account page.

jinoosss (approved):

PR #362, opened 2024-02-01
fix: [ADN-409] Fix Cannot Sign Transaction popup

jinoosss:


packages/adena-extension/src/common/provider/wallet/wallet-provider.tsx
```diff
@@ -75,6 +75,9 @@ export const WalletProvider: React.FC<React.PropsWithChildren<unknown>> = ({ chi
     setWalletStatus('LOADING');
     try {
       const wallet = await walletService.loadWallet();
+      console.log(wallet);
```
skqksh, line 78:
```suggestion
```
remove console

skqksh (approved):

PR #396, opened 2024-02-13
fix: [ADN-465] Fix Add Establish

jinoosss:


packages/adena-extension/src/router/popup/navigation/index.tsx
skqksh, line 1: FYI. you can use `useRecoilValue` instead of `useRecoilState`

```
  const walletState = useRecoilValue(WalletState.state);
  const failedNetwork = useRecoilValue(CommonState.failedNetwork);```

skqksh (approved):

PR #402, opened 2024-02-22
feat: Encrypt address book data

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
- Encrypts address books that are stored in plain text in Storage.
- Adds a loading screen from the address book screen due to encryption.
- Resolves:
    - Adena Security Evaluation Feb 2024 - LOW-002

### DIFF
--- a/packages/adena-extension/src/hooks/use-address-book.tsx
+++ b/packages/adena-extension/src/hooks/use-address-book.tsx
@@ -1,0 +1,84 @@
+import { useRecoilState } from 'recoil';
+
+import { useAdenaContext } from './use-context';
+import { WalletState } from '@states';
+import { useCurrentAccount } from './use-current-account';
+import { AddressBookItem } from '@repositories/wallet';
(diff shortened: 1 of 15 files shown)


packages/adena-extension/src/hooks/use-address-book.tsx
```diff
@@ -0,0 +1,84 @@
...
+        .then(() => true)
+        .catch(() => false),
+    );
+  };
+
+  const initAddressBook = async (): Promise<boolean> => {
+    return _update();
+  };
+
+  const _update = async (callback?: () => Promise<boolean>): Promise<boolean> => {
```
skqksh, line 61: Would better remove callback param

jinoosss: I thought this would be a good way to handle common logic before and after executing a function.
Can you share any suggestions on how to do this ?

PR #471, opened 2024-04-09
fix: Fix bugs when changing network

jinoosss:
### What type of PR is this?

- bug

### What this PR does:
- Change to initialize data for loading conditions when the network changes.


dongwon8247 (approved):
Well done

PR #473, opened 2024-04-09
fix: Fix Ui issues

jinoosss:
### What type of PR is this?

- bug
- style

### What this PR does:
- Fixed a scrolling UI issue.
- Fixed minor UI issues.


dongwon8247 (changes requested):
Checked all UI fixes except the ledger one (d04b88acdf39ae159bcea8168a392a7f074afbbd, let's check this on QA), and all look great but one minor issue - when you pop-up window in pages with logos and balances, it seems to be taking longer and showing unloaded image & data as you can see below. Can you check on this?

image
image

jinoosss:
> Checked all UI fixes except the ledger one ([d04b88a](https://github.com/onbloc/adena-wallet/commit/d04b88acdf39ae159bcea8168a392a7f074afbbd), let's check this on QA), and all look great but one minor issue - when you pop-up window from pages with logos and balances, it seems to be taking longer and showing unloaded image & data as you can see below. Can you check on this?

It turned out that the problem was caused by loading the transaction history first before fetching the token information.
I changed it to work after the token information is loaded.

dongwon8247 (approved):
LGTM!

PR #474, opened 2024-04-09
fix: Fix injection and pop window collision

jinoosss:
### What type of PR is this?

- bug

### What this PR does:
- Changed to remove popups except for the injection popup when running an injection with a tab open in a popup window.
     - This resolves an error that occurs when running an injection popup, if one exists.


dongwon8247 (approved):
Well done 💯

Issue #480, opened 2024-04-12
Local development network


kazai777:
Hi, when I change network to `local development network` the first launch does it correctly and when I leave adena and go back a second time it loads indefinitely, like in the photo below. I would also like to know how to add tokens to the `local development network`, because I tried to add my adena address to the `genesis_balance.txt` file of `gno.land/genesis` but the tokens are not loaded on adena.
![Screenshot 2024-04-12 at 21 09 22](https://github.com/onbloc/adena-wallet/assets/149690535/58d786b0-01e0-4ef5-9838-945e543c0d8b)

dongwon8247:
@kazai777 Thank you for the report.

We will look into the infinite loading problem when using 'local'.

For adding tokens in 'local', Adena only supports a filter/search function for adding tokens currently, not supporting adding it manually. We're working on adding this feature in the next release so that you will be able to manually add your tokens in any network with a 'path'.

Will open this issue until we add this feature. Thank you!

dongwon8247:
@kazai777 The infinite loading problem has been resolved in the recent hotfix (https://github.com/onbloc/adena-wallet/releases/tag/v1.10.1). The feature to add GRC20 manually will be added later.

dongwon8247:
Closing as completed

PR #483, opened 2024-04-15
fix: Fix Infinity loading

jinoosss:
### What type of PR is this?

- bug

### What this PR does:
- Fixes a bug in the wallet that causes persistent loading after a network change.
- Change network state variables to be managed by React Query.
- Reference Issue: https://github.com/onbloc/adena-wallet/issues/480


dongwon8247 (approved):
Tested and it works well.

It only happened in MacOS, but seems to we resolved by this PR. Well done!

PR #502, opened 2024-04-29
feat: Use a separate tm2-js-client per network

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
- Due to changes in the signing process, we use a separate `tm2-js-client` for each network.
- The `test3` network uses an `tm2-js-client@1.1.7`.


jinoosss:
When `test3` network support ends, 
we will need to remove dependencies on older versions of `tm2-js-client`. 

cc) @dongwon8247

zivkovicmilos (approved):

PR #529, opened 2024-07-14
chore: add faucet API to test4 network

jinoosss:
### What type of PR is this?

- chore

### What this PR does:
- Add the faucet API information to support Faucet on the test4 network.


dongwon8247 (approved):
TODO after releasing new version: remove `test3`

PR #571, opened 2024-09-10
feat: Naming service

thanhngoc541:
### What type of PR is this?
Feature

### What this PR does:
Adds a naming service for wallets. Users can now register a domain name, which acts as a shortcut for their wallet address, at https://gno-naming-service.vercel.app/. The registered domain name will be displayed in the wallet and can be used in place of the wallet address.

This feature is currently available on the test4 network and will be rolled out to other networks soon.

### DIFF
--- a/packages/adena-extension/.storybook/preview.tsx
+++ b/packages/adena-extension/.storybook/preview.tsx
@@ -10,7 +10,7 @@ const extensionViewport = {
   name: 'Adena Extension',
   styles: {
     width: '360px',
-    height: '540px',
+    height: '566px',
   },
 };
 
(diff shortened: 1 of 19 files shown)


jinoosss:
Thank you for creating a PR 🙏 

However, before i can review this PR, i need to discuss whether we should support this feature.

Adena aims for formal, standardised feature support for Gno.

I don't think you should create dependencies on services that are managed by a specific team or individual. 

This PR seems to be for feature support for `gno.land/r/varmeta/demo/v1/domain/resolver`, 
should we support this specific feature?

jinoosss:
cc @dongwon8247

dongwon8247:
Closing

PR #582, opened 2024-09-24
feat: add an Import Wallet flow to support multiple mnemonics

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
- Added a new “Import Wallet” flow that supports the entry of multiple mnemonic phrases.
- Updated UI components to handle multiple mnemonics during wallet import.
- Adjusted the state management to store multiple mnemonics and corresponding wallet addresses.


dongwon8247 (approved):

packages/adena-extension/src/common/validation/validation-wallet.ts
```diff
@@ -0,0 +1,8 @@
+export function isSeedPhraseString(seedPhrase: string, length: number): boolean {
```
zivkovicmilos, line 1: Just curious, do we ever check for the validity of the mnemonic?
Obviously not here, but in general

packages/adena-extension/src/hooks/web/use-account-import-screen.ts
```diff
@@ -101,61 +146,154 @@ const useAccountImportScreen = ({ wallet }: { wallet: Wallet }): UseAccountImpor
...
         navigate(RoutePath.WebQuestionnaire, {
           state: {
             callbackPath: RoutePath.WebAccountImport,
           },
         });
       }
-    } else if (step === 'SET_PRIVATE_KEY') {
-      const result = await makePrivateKeyAccountAndKeyring();
-      if (!result) {
+    } else if (step === 'SET_MNEMONIC') {
```
zivkovicmilos, line 178: I wish I had confidence in life like @jinoosss has in typing out these constants correctly every time 😄

packages/adena-extension/src/common/validation/validation-wallet.ts
jinoosss: Normally, when creating a mnemonic, it would be checked if it is a valid mnemonic. 
- Whether it is included in a BIP39 compliant managed wordlist.
- Number of words, EnglishMnemonic in cosmjs supports 12, 15, 18, 21, 24.

It will be validated when creating the wallet for the tm2-js-client.
https://github.com/gnolang/tm2-js-client/blob/main/src/wallet/utility/utility.ts#L56

The function on this line is a function that checks the number of words the user has entered,
Now that I think about it, I should change the function name to indicate that it checks for word length.

packages/adena-extension/src/hooks/web/use-account-import-screen.ts
jinoosss: Oh, Thanks for the good point!
I was using the string literal type.
https://github.com/onbloc/adena-wallet/pull/582/files/c3d144247b0a3feeb8c0dac741bc6bc36183ec2c#diff-fd678e1a97fbf36bebf013ce3147d3f291a7ebabed8523e436e64c90c102c021R60

I should have used the enum type, which is much clearer.

PR #591, opened 2024-09-26
chore: update gno docs link url

jinoosss:
### What type of PR is this?

- chore

### What this PR does:
- Change the link URL to the Gno CLI documentation. 
(https://docs.gno.land/gno-tooling/cli/gnokey/working-with-key-pairs)

### DIFF
--- a/packages/adena-extension/src/common/constants/resource.constant.ts
+++ b/packages/adena-extension/src/common/constants/resource.constant.ts
@@ -6,5 +6,5 @@ export const ADENA_SETUP_AIRGAP_HELP_PAGE =
 export const HARDWARE_WALLET_LEARN_PAGE =
   'https://docs.adena.app/user-guide/sign-in/connect-hardware-wallet' as const;
 export const GNO_CLI_HELP_PAGE =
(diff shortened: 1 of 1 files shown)


dongwon8247 (approved):

packages/adena-extension/src/common/constants/resource.constant.ts
```diff
@@ -6,5 +6,5 @@ export const ADENA_SETUP_AIRGAP_HELP_PAGE =
 export const HARDWARE_WALLET_LEARN_PAGE =
   'https://docs.adena.app/user-guide/sign-in/connect-hardware-wallet' as const;
 export const GNO_CLI_HELP_PAGE =
-  'https://docs.gno.land/getting-started/local-setup/working-with-key-pairs' as const;
+  'https://docs.gno.land/gno-tooling/cli/gnokey/working-with-key-pairs' as const;
```
zivkovicmilos, line 9: It can be sketchy to hardcode something like this, but just to keep in mind we should probably point it to a shortened URL we can change

jinoosss: I'll try to figure out a way to make it manageable.
Environment variables, static files, remotes, etc.

PR #604, opened 2024-10-15
feat: add a feature to manually add custom tokens

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
Change to allow GRC20 tokens to be added manually.

Call the RPC directly for the presence of tokens, rather than a list of tokens fetched from the tx-indexer.
- To determine if Adena is a token, we parse the result of GRC20's Render function for information.


dongwon8247 (changes requested):
Locally tested, mostly works well, but minor fixes needed:

1. When a token path is long, such as `gno.land/r/g146gxysfx24t3nar4z4yzkca2g8u9vl8xlpmtq0/tuna`, it doesn't detect the token.
2. Some tokens, such as `gno.land/r/onbloc/foo`, had a 5-6 second loading time when adding them. 
3. When I manually add a token that I've already added, it still adds even if it's already added. If it's been added already, then can we show "Already added" in the error message?
4. Can you make the three small loading bars blink? like the main page's loading bars.
5. We talked about Adena automatically detecting tokens and adding them when there's balance. Do you want to work on this on a separate PR? up to you.

jinoosss:
@dongwon8247 ,
Thanks for the review.
I'll take a look at the contents.

And I think detecting and adding tokens with balance should be a separate task.

dongwon8247 (approved):
👍

PR #609, opened 2024-10-22
feat: add a memo field

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
- When sending a token, we provide the ability to send a note along with it.
- Provides the ability to edit or check notes when executing the Transaction Broadcast request method in the DApp.


dongwon8247 (approved):
locally tested and it works great!

PR #631, opened 2024-11-10
fix: fix infinite loading in custom network environments

jinoosss:
### What type of PR is this?

- bug

### What this PR does:

Fixes an error that causes infinite loading on non-official networks.
- Fix conditional statements in guard logic in the function that initializes token information.


dongwon8247 (approved):

dongwon8247:
hotfix for v1.13.1

PR #671, opened 2025-01-27
fix: check the source when handling events

jinoosss:
### What type of PR is this?

- fix

### What this PR does:

- Restrict event messages to be delivered only to the source on the current page.
The trusted domains that can then execute Adena commands are managed by an in-service feature in Adena.

- Check the event source address in the event message listener function.
- When sending an event message, specify the target origin.


dongwon8247 (approved):
LGTM!

PR #684, opened 2025-01-30
feat: add GRC20 token validation logic

jinoosss:
### What this PR does:
- When adding GRC20 tokens, add validation logic.
    - Configure the metadata using functions in the Gno.land RPC via the token path fetched by the API.
    - Configure metadata with actual chain data.


dongwon8247 (approved):
LGTM!

PR #738, opened 2025-05-04
fix: remove questionnaire step in wallet export

jinoosss:
### What type of PR is this?

- bug

### What this PR does:
- Remove questionnaire step in wallet export.


dongwon8247 (approved):
@zivkovicmilos 👀

PR #741, opened 2025-05-19
feat: query gas price from chain

jinoosss:
### What type of PR is this?

- feature

### What this PR does:
This PR implements direct gas price querying from the chain instead of using the indexer. The changes include updating the gas price calculation logic and simplifying the related interfaces.

## Changes
- Implement `getGasPrice` method in `GnoProvider` to query gas price directly from chain
- Update gas price constants and calculation logic
- Simplify gas price related interfaces and remove unnecessary complexity
- Remove indexer-based gas price fetching

### Gas Price Query
- Added new ABCI query endpoint for gas price
- Implemented gas price parsing from chain response
- Updated gas price calculation with new base values

### Interface Changes
- Simplified `ITransactionGasRepository` interface
- Removed `GasPriceTierInfo` type

### DIFF
--- a/packages/adena-extension/src/common/constants/gas.constant.ts
+++ b/packages/adena-extension/src/common/constants/gas.constant.ts
@@ -17,8 +17,8 @@ export const DEFAULT_GAS_PRICE_STEP: Record<NetworkFeeSettingType, number> = {
 } as const;
 
 export const DEFAULT_GAS_PRICE_RATE: Record<NetworkFeeSettingType, number> = {
-  FAST: 4,
-  AVERAGE: 2.5,
+  FAST: 1.2,
+  AVERAGE: 1.1,
   SLOW: 1,
 } as const;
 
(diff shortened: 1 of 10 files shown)


packages/adena-extension/src/common/constants/gas.constant.ts
```diff
@@ -17,8 +17,8 @@ export const DEFAULT_GAS_PRICE_STEP: Record<NetworkFeeSettingType, number> = {
 } as const;
 
 export const DEFAULT_GAS_PRICE_RATE: Record<NetworkFeeSettingType, number> = {
-  FAST: 4,
-  AVERAGE: 2.5,
+  FAST: 1.2,
+  AVERAGE: 1.1,
```
dongwon8247, line 21: Do you think this will be sufficient? (I get the reason because it's from the chain, which is more correct and reliable)

dongwon8247: Just tested, transactions don't go through with the suggested average value. It goes through with the fast option.

We need to find the right multiplier that's just enough to make a transaction go through, which we should use for the average option.

dongwon8247: Also, currently, users can send transactions even if the transaction fee is too little to send a transaction. 
1. We need to know the lowest, minimum value that can make transactions go through
2. If the Network Fee is lower than that, then we should error out

jinoosss: That's right, 
sets the gasUsed data from simulating the current transaction to the minimum value.
I'll have to find the right weighting values through testing.

The handling of values less than the minimum will be covered in the migration task.

PR #781, opened 2025-10-20
fix: resolve inactive account login failure

tfrg:
### What type of PR is this?
- fix

#### Problem
Inactive accounts (wallets with no on-chain transaction history) were unable to log in due to the `address` field being returned as an empty string (`""`), causing address validation to fail during the login process.

#### Reproduction Steps
1. Attempt to log in with a new wallet address that has no on-chain history
2. `getAccountInfo()` returns `{ address: "", status: 'IN_ACTIVE', ... }`
3. Login validation fails because `address` is empty
4. User cannot proceed with login ❌

##### Key Changes
1. **Return actual address parameter for inactive accounts**
   ```typescript
   // Before
   return {
     address: "",
     status: 'IN_ACTIVE',
     ...
   };
   
   // After
   return {
     address,  // ✅ Actual wallet address from parameter
     status: 'IN_ACTIVE',
     ...
   };

### DIFF
--- a/packages/adena-extension/src/common/provider/gno/gno-provider.ts
+++ b/packages/adena-extension/src/common/provider/gno/gno-provider.ts
@@ -91,39 +91,46 @@ export class GnoProvider extends GnoJSONRPCProvider {
     address: string,
     height?: number | undefined,
   ): Promise<AccountInfo | null> {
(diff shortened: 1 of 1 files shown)


dongwon8247 (approved):
Verified that it works fine now.

PR #812, opened 2026-04-15
ADN-749: add ChainRegistry and TokenRegistry to adena-module

HeesungB:
## Summary

Foundation work in `adena-module` for Phase 1 multichain (AtomOne) support. Pure addition — no existing code changed, no regression risk.

- Add `ChainRegistry` — `chainType: 'gno' | 'cosmos'` discriminated union, `chainGroup` (Permission isolation) / `isMainnet` fields, `listByGroup` / `getDefault` methods
- Add `TokenRegistry` — chain-scoped lookup, origin discriminated union (gno-native / cosmos-native / cosmos-ibc / cosmos-factory)
- Chain entries: 5 Gno (GNOLAND1 mainnet, STAGING, TEST12, DEV, GNOSWAP) + 1 AtomOne (ATOMONE_1 mainnet)
- Token entries: UGNOT, UATONE, UPHOTON
- 3 icon SVGs
- Singleton `defaultChainRegistry` / `defaultTokenRegistry` + barrel export

## Test plan

- [ ] `cd packages/adena-module && npm test`
- [ ] `npx tsc --noEmit`
- [ ] `defaultChainRegistry.getDefault('gno')?.id === 'gnoland1'`
- [ ] `defaultChainRegistry.getDefault('atomone')?.id === 'atomone-1'`
- [ ] `defaultTokenRegistry.list('atomone-1')` → `[UATONE, UPHOTON]`

## Related

Follow-up: `feature/ADN-749-2` (bech32 prefix parameterization)


packages/adena-module/src/token-registry/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from './types';
+export * from './registry';
+export * from './entries';
+
+import { TokenRegistryImpl } from './registry';
+import { ALL_TOKENS } from './entries';
+
+export const defaultTokenRegistry = new TokenRegistryImpl();
+
+ALL_TOKENS.forEach((token) => defaultTokenRegistry.register(token));
```
jinoosss: Instead of managing instances globally within the module, it seems better to just provide the module and let the app manage it via state.

packages/adena-module/src/chain-registry/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export * from './entries/atomone';
+
+import { ChainRegistryImpl } from './registry';
+import { GNO_CHAINS } from './entries/gno';
+import { ATOMONE_CHAINS } from './entries/atomone';
+
+export const defaultChainRegistry = new ChainRegistryImpl();
+
+const ALL_CHAINS = [...GNO_CHAINS, ...ATOMONE_CHAINS];
+ALL_CHAINS.forEach((chain) => defaultChainRegistry.register(chain));
```
jinoosss: Instead of managing instances globally within the module, it seems better to just provide the module and let the app manage it via state.

jinoosss (commented):
LGTM 👍 . 
I left a comment.

packages/adena-module/src/token-registry/index.ts
HeesungB: I have made the changes you requested. Thank you.

packages/adena-module/src/chain-registry/index.ts
HeesungB: I have made the changes you requested. Thank you.

packages/adena-module/src/chain-registry/entries/gno.ts
```diff
@@ -0,0 +1,95 @@
...
+  linkUrl: 'http://127.0.0.1:3000',
+};
+
+export const GNO_GNOSWAP: GnoNetworkProfile = {
+  ...GNO_PROFILE_BASE,
+  id: 'gnoswap',
+  chainId: 'dev.gnoswap',
+  displayName: 'GnoSwap Dev',
+  isMainnet: false,
+  nativeTokenId: 'gnoswap:ugnot',
```
jinoosss:
```suggestion
  id: 'dev.gnoswap',
  chainId: 'dev.gnoswap',
  displayName: 'GnoSwap Dev',
  isMainnet: false,
  nativeTokenId: 'dev.gnoswap:ugnot',
```

packages/adena-module/src/chain-registry/types.ts
```diff
@@ -0,0 +1,111 @@
...
+  listNetworkProfilesByChain(chainGroup: string): NetworkProfile[];
+  getDefaultNetworkProfile(chainGroup: string): NetworkProfile | undefined;
+
+  // Backward-compatible NetworkProfile API (keeps existing register/get/list/listByGroup/getDefault)
+  register(profile: NetworkProfile): void;
+  get(id: string): NetworkProfile | undefined;
+  list(): NetworkProfile[];
+  listByGroup(chainGroup: string): NetworkProfile[];
+  getDefault(chainGroup: string): NetworkProfile | undefined;
+}
```
jinoosss, line 111: Is there a reason why the response is returned as `undefined` instead of `null`?

jinoosss (commented):
I've left a few comments.

packages/adena-module/src/chain-registry/types.ts
HeesungB: In TS/JS, "lookup miss = absence of a value" is idiomatically expressed as undefined. Map.get(), Array.find(), and optional property access all return undefined to signal "not found", whereas null is closer to "explicitly assigned as empty".
Since this registry uses Map internally, exposing the returned value as-is is the most natural choice — wrapping it with ?? null would only add a meaningless conversion layer.
Mixing null and undefined also widens the return type to T | null | undefined and increases defensive checks at call sites, so I unified everything to undefined. It also composes more naturally with optional chaining (?.), nullish coalescing (??), and default parameter syntax.

In short, since a "lookup miss" represents the absence of a value, I went with undefined following the JS/TS convention.

That said, I've noticed some places in the current project that explicitly use null — is that an intentional convention, or is there another reason behind it?

jinoosss: I understand your point.

There were several reasons for using `null` previously.
When used in interfaces and implementations, the goal was to create a generic interface, with implementations aiming to conform to that interface’s specifications.
When used in existing variables, the purpose was to distinguish between `undefined` and `null`. I believe this reflects the difference between a value not existing and a value being absent.

jinoosss (approved):

packages/adena-module/src/chain-registry/types.ts
jinoosss: The current implementation is acceptable.

PR #814, opened 2026-04-17
ADN-749: add resolveAddress cache to optimize multichain address derivation

HeesungB:
## Summary

Avoid redundant crypto work when deriving addresses for the same `Account` across multiple prefixes in multichain balance/display loops.

- Add `resolveAddress(prefix)` to the `Account` interface and all implementations — returns a cached address per prefix, deriving and caching on first call
- Back the cache with a module-level `WeakMap<Account, …>` in `src/wallet/account/address-cache.ts`. This keeps `Account` instances fully immutable (required when they are held in Recoil atoms that deep-freeze their values); the `WeakMap` also auto-collects entries when the `Account` is GC'd
- `getAddress` remains uncached for callers that need fresh derivation; `resolveAddress` is preferred for multi-chain loops
- Update `wallet.hasAddress` to use `resolveAddress` for cache benefit
- Remove `defaultChainRegistry` / `defaultTokenRegistry` singletons — let the app own registry instances
- Fix storybook mock: add `resolveAddress` to the `Account` mock (TS2741)

## Test plan

- [ ] `cd packages/adena-module && npm test` — new `account.spec.ts` (cache hit/miss, prefix independence, WeakMap isolation between instances, cache miss after `toData` / `fromData`)
- [ ] `cd packages/adena-extension && npx tsc --noEmit`

## Related

Base: `feature/ADN-749-2` (bech32 prefix parameterization)


packages/adena-module/src/wallet/account/address-cache.ts
```diff
@@ -0,0 +1,26 @@
+const addressCache = new WeakMap<object, Map<string, string>>();
```
jinoosss: It seems better to manage this cache data as part of the app's state data.

PR #818, opened 2026-04-17
ADN-750: add Keyring.signRaw and extract signGnoDocument helper

HeesungB:
## Summary

Phase 2 of AtomOne multichain support. Splits the keyring signing API so that chain-specific serialization (Gno JSON, Cosmos AMINO/DIRECT) is decoupled from the key-holding primitive. Pure refactor — Gno signatures remain bit-equal to the pre-change tm2 pipeline.

- Add `signRaw(bytes, opts?)` to the `Keyring` interface — returns raw 64-byte `r||s` ECDSA signature
- Implement for `HDWalletKeyring`, `PrivateKeyKeyring`, `Web3AuthKeyring` via shared `signRawWithPrivateKey`
- Stub `LedgerKeyring` (Phase 7), `AddressKeyring` (AIRGAP), `MultisigKeyring` with throw-and-explain comments
- Extract Gno signing pipeline into `signGnoDocument(provider, document, keyring, opts?)` so future Cosmos pipelines (AMINO / DIRECT) can share the same primitive
- Route HD / PK / Web3Auth `.sign()` through the new helper; remove dead `signByWallet` private methods
- Preserve bit-equal signatures via a golden hex vector regression test in `wallet-sign.spec.ts`

## Test plan

- [x] `cd packages/adena-module && npx jest src/wallet/keyring/sign-raw.spec.ts src/wallet/keyring/sign-gno-document.spec.ts src/wallet/wallet-sign.spec.ts` → 20/20 passing
- [x] `npx tsc --noEmit` → clean on src/
- [x] Parity test: `Tx.encode` output identical between legacy tm2 path and new `signGnoDocument` path
- [x] Golden signature hex check for fixed mnemonic + document
- [ ] Manual smoke on dev build: HD / PrivateKey / Web3Auth account signs a `/vm.m_call` tx and broadcasts successfully
- [ ] Manual smoke: Ledger account still signs via existing path (unchanged in this PR)
- [ ] Manual smoke: Multisig request UI unchanged

## Related

Plan: `plans/ADN-750/plan.md`
Spec: `plans/adena-atomone-support/phase-02-keyring-signraw.md`
Follow-up: Phase 3 (Cosmos AMINO pipeline) will consume `signRaw` from this PR

PR #821, opened 2026-04-22
ADN-751: add Cosmos AMINO send flow for AtomOne (Phase 3)

HeesungB:
## Summary

Phase 3 of AtomOne multichain support. End-to-end `MsgSend` pipeline — UI input → AMINO signing → TxRaw serialization → AtomOne LCD broadcast — built on top of Phase 2's `Keyring.signRaw` (#818) and Phase 1's ChainRegistry / CosmosLcdProvider / balances UI (#820). Gno send/sign paths are strictly untouched.

### adena-module

- New `cosmos/` subtree: shared types, `CosmosProvider` interface (Gno-style DI so adena-module stays HTTP-free), MsgSend codec (amino ⇄ proto) with a codec registry, `makeTxRaw` (pubkey compression, `AuthInfo`/`TxBody`/`TxRaw` encoding)
- AMINO (`signCosmosAmino`) + DIRECT (`signCosmosDirect`) pipelines, dispatched by `signCosmos` based on `CosmosSignMode`
- `validateCosmosAddress` — bech32 prefix mismatch guard (blocks `gno1...` on AtomOne)
- `Wallet.signCosmosByAccountId` / `broadcastCosmosTx` — dedicated methods so Gno callers don't see union-typed returns

### adena-extension

- `TransactionService.signCosmos` / `broadcastCosmos` — resolve profile + preferred `CosmosSignMode` via ChainRegistry, delegate to Wallet, invalidate provider dedupe cache after broadcast
- `CosmosLcdProvider.getAccount` — `/cosmos/auth/v1beta1/accounts/{addr}` with `base_account` unwrap; empty-string `account_number`/`sequence` coerced to `'0'` so first-send accounts don't hit `BigInt('')`
- `CosmosLcdProvider.broadcastTx` — throws with `code` + `raw_log` on non-zero code so LCD errors surface to the Failure screen
- `transfer-input`: Cosmos prefix validation; AddressInput `maxLength`/placeholder become props (default Gno 40, Cosmos 48)
- `use-token-balance`: separate Cosmos query — keyed by `(accountId, atomoneNetworkId)`, 10s refetch, `retry: 1`, never blocks Gno rendering
- `transfer-summary` Cosmos branch: builds `MsgSend`, Phase 3 MVP hardcoded fee `2000uphoton` / gas `200000`, Mintscan link via `profile.linkUrl`
- `token-details`: Multisig × Cosmos permanently unsupported (tooltip)
- Remove Phase 1 `readOnly` guard — Cosmos tokens are now sendable
- Four `TODO(Phase 6)` markers on hardcoded-fee sites for the feemarket migration (grep-able)

### Tests

- adena-module: AMINO/DIRECT golden vectors, TxRaw pubkey compression, msg-send round-trip, cosmos-address prefix match, Gno `.sign()` bit-equal parity with pre-Phase-2 output
- adena-extension: `cosmos-lcd-provider.spec.ts` (base_account unwrap, empty-string coerce, broadcast throw on non-zero code, invalidate clears dedupe), `cosmos-query-client.spec.ts` HTTP contract

## Test plan

- [ ] `cd packages/adena-module && npx jest` pass
- [ ] `cd packages/adena-extension && npm test` pass
- [ ] `npx tsc --noEmit` clean, `yarn build` green
- [ ] Manual: send `ATONE` on AtomOne — tx appears on Mintscan
- [ ] Manual: send `PHOTON` — balance updates after invalidate
- [ ] Manual: paste `gno1...` on Cosmos send form — "Invalid Address" before Next
- [ ] Manual: zero `PHOTON` balance — node's "insufficient fees" `raw_log` shown on the Failure screen
- [ ] Manual: AtomOne ⇄ Gno network switch — per-chain balance lifecycles stay independent
- [ ] Manual: Multisig × Cosmos — Send disabled with expected tooltip
- [ ] Manual: Ledger × Cosmos — rejected with Phase 7 message, no crash
- [ ] Manual: Gno GNOT / GRC20 transfer unchanged

## Related

- Base: \`feature/ADN-743\`
- Depends on: #820 (ADN-749), #818 (ADN-750)
- Spec: \`plans/cosmos-support/phase-03-cosmos-amino-msgsend.md\`
- Plan: \`packages/adena-extension/plans/ADN-751/{plan,implementation}.md\`
- Follow-up:
  - Phase 6 — feemarket dynamic fee (replaces the four \`TODO(Phase 6)\`)
  - Phase 7 — Ledger Cosmos signing
  - Cosmos endpoint failover (array shape already preserved)

### DIFF
--- a/packages/adena-extension/src/pages/popup/wallet/transfer-input/index.tsx
+++ b/packages/adena-extension/src/pages/popup/wallet/transfer-input/index.tsx
@@ -1,3 +1,4 @@
+import { validateCosmosAddress } from 'adena-module';
 import BigNumber from 'bignumber.js';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
@@ -5,6 +6,7 @@ import { isNativeTokenModel } from '@common/validation/validation-token';
 import TransferInput from '@components/pages/transfer-input/transfer-input/transfer-input';
 import { useAddressBookInput } from '@hooks/use-address-book-input';
 import { useBalanceInput } from '@hooks/use-balance-input';
+import { useAdenaContext } from '@hooks/use-context';
 import { useCurrentAccount } from '@hooks/use-current-account';
 import useHistoryData from '@hooks/use-history-data';
 import { RoutePath } from '@types';
@@ -44,7 +46,15 @@ const TransferInputContainer: React.FC = () => {
   } = useSessionParams<RoutePath.TransferInput>();
   const [isTokenSearch, setIsTokenSearch] = useState(params?.isTokenSearch === true);
   const [tokenMetainfo, setTokenMetainfo] = useState<TokenModel | undefined>(params?.tokenBalance);
-  const addressBookInput = useAddressBookInput();
+  const { chainRegistry } = useAdenaContext();
+  // For Cosmos-native tokens the recipient/address-book must use the target
+  // chain's bech32 prefix (e.g. 'atone') instead of the current Gno network's.
+  const addressPrefixOverride = useMemo(() => {
+    if (tokenMetainfo?.type !== 'cosmos-native') return undefined;
+    const chain = chainRegistry.getChainByChainId(tokenMetainfo.networkId);
+    return chain?.chainType === 'cosmos' ? chain.bech32Prefix : undefined;
+  }, [tokenMetainfo, chainRegistry]);
+  const addressBookInput = useAddressBookInput(addressPrefixOverride);
   const balanceInput = useBalanceInput(tokenMetainfo);
   const { currentAccount } = useCurrentAccount();
   const { getHistoryData, setHistoryData } = useHistoryData<HistoryData>();
@@ -102,6 +112,22 @@ const TransferInputContainer: React.FC = () => {
     goBack();
   }, [isTokenSearch]);
 
+  const validateCosmosPrefixIfNeeded = useCallback((): boolean => {
+    if (!tokenMetainfo || tokenMetainfo.type !== 'cosmos-native') {
+      return true;
+    }
+    const chain = chainRegistry.getChainByChainId(tokenMetainfo.networkId);
+    if (!chain || chain.chainType !== 'cosmos') {
+      return true;
+    }
+    const ok = validateCosmosAddress(addressBookInput.resultAddress, chain.bech32Prefix);
+    if (!ok) {
+      // Reuse the generic "Invalid Address" UI state from addressBookInput.
+      addressBookInput.validateAddressBookInput();
+    }
+    return ok;
+  }, [tokenMetainfo, chainRegistry, addressBookInput]);
+
   const onClickNext = useCallback(async () => {
     if (!isNext) {
       return;
@@ -111,6 +137,7 @@ const TransferInputContainer: React.FC = () => {
     }
     const validAddress =
       addressBookInput.validateAddressBookInput() &&
+      validateCosmosPrefixIfNeeded() &&
       (isNativeTokenModel(tokenMetainfo) || (await addressBookInput.validateEqualAddress()));
     const validBalance = balanceInput.validateBalanceInput();
     if (validAddress && validBalance) {
@@ -129,7 +156,7 @@ const TransferInputContainer: React.FC = () => {
         },
       });
     }
-  }, [addressBookInput, balanceInput, isNext]);
+  }, [addressBookInput, balanceInput, isNext, validateCosmosPrefixIfNeeded]);
 
   useEffect(() => {
     if (isLoadingSessionState) {
--- a/packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
+++ b/packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
@@ -314,15 +314,13 @@ const TransferSummaryContainer: React.FC = () => {
     }
 
     // Cosmos AMINO path (Phase 3) — skips Gno createDocument/gas-simulate flow.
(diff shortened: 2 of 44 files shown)


dongwon8247 (approved):

packages/adena-extension/src/pages/popup/wallet/transfer-input/index.tsx
```diff
@@ -44,7 +46,15 @@ const TransferInputContainer: React.FC = () => {
...
   const [tokenMetainfo, setTokenMetainfo] = useState<TokenModel | undefined>(params?.tokenBalance);
-  const addressBookInput = useAddressBookInput();
+  const { chainRegistry } = useAdenaContext();
+  // For Cosmos-native tokens the recipient/address-book must use the target
+  // chain's bech32 prefix (e.g. 'atone') instead of the current Gno network's.
+  const addressPrefixOverride = useMemo(() => {
+    if (tokenMetainfo?.type !== 'cosmos-native') return undefined;
+    const chain = chainRegistry.getChainByChainId(tokenMetainfo.networkId);
+    return chain?.chainType === 'cosmos' ? chain.bech32Prefix : undefined;
+  }, [tokenMetainfo, chainRegistry]);
```
jinoosss, line 56: Is there a reason why types other than `cosmos` return `undefined`?

packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
```diff
@@ -221,11 +245,84 @@ const TransferSummaryContainer: React.FC = () => {
...
   };
 
+  // Cosmos (Phase 3): MVP hardcodes fee at 2000uphoton / gas 200000.
+  // Node's minGasPrice is 0.01 uphoton/gas → 0.01 * 200000 = 2000uphoton required.
+  // TODO(Phase 6): replace with feemarket dynamic estimation; move constant to
+  //                common/constants/tx.constant.ts and read denom from chain profile.
+  const COSMOS_DEFAULT_FEE = {
+    amount: [{ denom: 'uphoton', amount: '2000' }],
+    gas: '200000',
+  };
```
jinoosss, line 255: If you define a default gas fee, it seems best to manage it in a file that handles static variables.
https://github.com/onbloc/adena-wallet/blob/main/packages/adena-extension/src/common/constants/gas.constant.ts

Is there a standard for the default gas fee?

packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
```diff
@@ -407,8 +541,15 @@ const TransferSummaryContainer: React.FC = () => {
...
-          networkFee={networkFee}
+          // TEMP (Phase 3 MVP): Cosmos fee is hardcoded so the Gno gas-simulate
+          // hook stays in its loading state forever (enabled=false when document
+          // is null). Force loading=false and inject the fixed fee here to unblock
+          // the Send button.
+          // TODO(Phase 6): remove this override once feemarket estimation lands.
+          isLoadingNetworkFee={isCosmosToken ? false : useNetworkFeeReturn.isLoading}
+          networkFee={
+            isCosmosToken ? { amount: '0.002', denom: 'PHOTON' } : networkFee
+          }
```
jinoosss, line 552: Will this change once the Cosmos transaction simulation feature is implemented?

jinoosss (approved):
Overall, it looks fine.

It appears that the patterns `tokenMetainfo.type === ‘cosmos-native’` and `chain?.chainType === ‘cosmos’` are commonly used.
It seems best to define and manage these as utility functions.

There appear to be a few build errors.
- `hooks/use-current-account`
- `adena-module/src/utils/messages.ts`: `gas_wanted` type error

packages/adena-extension/src/pages/popup/wallet/transfer-input/index.tsx
HeesungB: `useAddressBookInput` defaults to the current Gno prefix (`g`), so the new parameter is shaped as an opt-in override — `undefined` for non-cosmos means "no override, fall back to the existing Gno default", which keeps the Gno path strictly untouched.

This branching is already gone in ADN-753: the hook now takes a `chainGroup` argument (`useAddressBookInput(tokenChainGroup)`) and the call site no longer special-cases cosmos.

packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
HeesungB: It's an intentional Phase 3 MVP placeholder kept in component scope to signal it's short-lived — the `TODO(Phase 6)` next to it flags the same concern. There's no "standard" value; it's just the minimum that satisfies the node's minGasPrice (0.01 uphoton × gas 200000 = 2000 uphoton).

I avoided putting it in `gas.constant.ts` because that file is shaped around Gno (`MINIMUM_GAS_PRICE`, `DEFAULT_GAS_USED`, etc.) and mixing a Cosmos-specific value would muddy the meaning.

ADN-754 removes the constant entirely and reads `chain.fee.fallbackFee` from ChainRegistry instead, so the "move to a constants file" question becomes moot.

packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
HeesungB: Yes — exactly that. ADN-754 introduces `useCosmosNetworkFee(cosmosDocument)`, which drives loading / error / amount from `simulateTx` + the node's min-gas-price query. This `isCosmosToken ? ...` override block is removed in that change.

PR #824, opened 2026-04-27
ADN-753: redesign Change Network UI and add AtomOne network mode foundation

HeesungB:
## Summary

- Redesign Change Network with chain-group sections and a testnet toggle; rewrite Add/Edit Custom Network for chainGroup-aware fields.
- Add `useChain` / `useNetworkProfile` hooks and a NetworkMetainfo→NetworkProfile mapper, then migrate `currentNetwork.*` readers to consume the typed profile.
- Introduce `ATOMONE_NETWORKS`, `NETWORK_MODE`, `CURRENT_ATOMONE_NETWORK_ID` storage keys; wire AtomOne metainfos and network mode through WalletProvider.
- Register `atomone-testnet-1` native tokens / icons and route `AdenaProvider` Cosmos path through the profile mapper (drops Phase 3's bech32-prefix bypass).
- Replace the Cosmos LCD provider with a profile-injected debounce wrapper; add a stateless `CosmosQueryClient` for REST LCD endpoints.
- Stabilize the loading overlay and balance queries during network transitions; hide tx history for Cosmos tokens with an unsupported notice.

## Test plan

- [ ] `cd packages/adena-module && npx jest`
- [ ] `cd packages/adena-extension && npm test`
- [ ] Manual: switch among Gno / AtomOne mainnet / testnet — header, balances, history all reflect the active chain
- [ ] Manual: Add / Edit Custom Network on both chain groups; chainGroup is preserved
- [ ] Manual: AtomOne testnet shows ATONE / PHOTON balances and icons

## Related

- Base: `feature/ADN-751` (depends on PR #821). Re-target to `feature/ADN-743` after #821 merges.
- Includes one small compat commit (`ADN-753: switch Cosmos address derivation to Account.getAddress`) that swaps `Account.resolveAddress` calls — left over from an early draft of the 749 cache API — for `Account.getAddress`, which is the method that actually shipped in `feature/ADN-743` via #820. Can be reverted once `feature/ADN-751` is rebased to fix the same references.
- Follow-up PRs in the same stack: ADN-754 / 755 / 756 / 760.

PR #827, opened 2026-04-27
ADN-756: expose window.adena.cosmos and Cosmos dApp connect/sign popups

HeesungB:
## Summary

- Expose `window.adena.cosmos` namespace with a Cosmos offline-signer wrapper; add Cosmos inject types, executor methods, and a JSON-safe SignDoc/Tx serializer.
- Add Cosmos background handlers (with top-level error guard) and the `SerializedCosmosKey` wire type; compress secp256k1 pubkey before exposing in `getKey`.
- New popup routes: `approve-establish-cosmos`, `approve-sign-cosmos`. Extend `AddEstablish` to a unified cross-protocol entry point and surface Cosmos connections in Connected Apps.
- Route Cosmos sign requests to the approve popup; open the unlock popup for `getKey` when the wallet is locked.
- Auto-switch network mode to match the target chain's mainnet/testnet; enable Ledger signing for Cosmos sign flows from dApps.
- Cosmos chainId / signDoc validation module + executor / handler / offline-signer unit tests.

## Test plan

- [ ] `cd packages/adena-module && npx jest`
- [ ] `cd packages/adena-extension && npm test`
- [ ] Manual: a Cosmos dApp calls `window.adena.cosmos.enable("atomone-testnet-1")` → Establish popup → Connected Apps shows it
- [ ] Manual: locked wallet × `getKey` → unlock popup, then key returned with compressed pubkey
- [ ] Manual: Cosmos signAmino / signDirect from dApp → approve-sign-cosmos popup → broadcast succeeds (hot account and Ledger)
- [ ] Manual: dApp targets a chainId on the inactive mainnet/testnet → wallet auto-switches and continues

## Related

- Base: `feature/ADN-755` (PR #826); re-target up the stack as ancestors merge.
- Includes one compat commit (`ADN-756: switch Cosmos getKey address derivation to Account.getAddress`) — same fix as ADN-753 PR #824 but for the Cosmos getKey handler and the locked-wallet getKey approve popup, plus the matching test mocks.
- Follow-up PR in the same stack: ADN-760.


packages/adena-extension/src/hooks/use-network.ts
```diff
@@ -154,13 +154,59 @@ export const useNetwork = (): NetworkResponse => {
...
+      if (justSwitched !== 'atomone') {
+        const atomonePair = atomoneNetworks.find(
+          (network) => !network.deleted && network.isMainnet === isMainnetTarget,
+        );
+        if (atomonePair && atomonePair.id !== currentAtomoneNetwork?.id) {
+          setCurrentAtomoneNetwork(atomonePair);
+          setSelectedProfileByChainGroup((prev) => ({ ...prev, atomone: atomonePair.id }));
+          await chainService.updateCurrentAtomoneNetworkId(atomonePair.id).catch(() => null);
+        }
+      }
```
jinoosss, line 190: In my opinion, it appears to use the same logic as `changeNetworkMode`.
https://github.com/onbloc/adena-wallet/blob/e0636f7868363c1bbc53b5cc039c37671ca2e4e8/packages/adena-extension/src/hooks/use-network.ts#L227

packages/adena-extension/src/services/wallet/wallet-establish.ts
```diff
@@ -1,29 +1,36 @@
...
+    const accountSites = await this.getEstablishedSitesBy(accountId);
+    if (chainId === undefined) {
+      // Legacy hostname-only check: any chainGroup connection counts. Used by
+      // header status indicators and other UI gates that pre-date the
+      // chainId-aware paths.
+      return accountSites.some((site) => site.hostname === hostname);
+    }
+    const siblingChainIds = this.resolveSiblingChainIds(chainId);
+    return accountSites.some(
+      (site) => site.hostname === hostname && siblingChainIds.includes(site.chainId),
```
jinoosss, line 33: This appears to be code intended for backward compatibility.
However, given that the `establish` policy has changed, why not block it strictly?  
I assume that if `establish` hasn't been completed in the dapp, it will request it anyway.

jinoosss (approved):
👍

packages/adena-extension/src/hooks/use-network.ts
HeesungB: Good catch — the cascade body does overlap with `changeNetworkMode`. The two semantic differences (early-return when `nextMode === mode`, and skipping the `justSwitched` chainGroup) are intentional: `syncModeFromTarget` runs after `changeNetwork` has already updated the user's pick, and re-processing that group through stale closures could overwrite the selection.

I'd like to keep the duplication as-is in this PR and follow up with a separate refactor that extracts the shared cascade body into a helper both functions call.

packages/adena-extension/src/services/wallet/wallet-establish.ts
HeesungB: This branch preserves the existing `main` behavior — `isEstablishedBy` on `main` has always been hostname-only with no chainId argument, and callers without a chainId (legacy single-chain dApps + UI status indicators) route to `establishService` (Gno) and fall back to that path.

The chainId-aware path is additive: new dApps that send `chainIds` go through the strict matching, and the new AtomOne service enforces chainId from day one. Tightening the legacy Gno path would break compatibility with existing single-chain dApp callers, so I'd prefer to keep it as-is here and track stricter enforcement as a separate migration ticket.

HeesungB:
Squash-merged into `feature/ADN-743` as 5 grouped commits (`417ba723`..`72b39817`). Closing in favor of those commits.

HeesungB:
Note: superseded by #831 — squashed merge of ADN-756 + ADN-760 into `feature/ADN-743`.

PR #836, opened 2026-04-29
security: address Findings #01–#06 from security review

HeesungB:
## Summary

Addresses six findings from the recent security review of the wallet
keyring, storage, and dApp connection paths.

- **Finding 01** — Replace AES-CBC with XChaCha20-Poly1305 AEAD for wallet
  serialization (`02c18984`). Adds the `v018` storage migration alongside
  per-user salt and AEAD nonce handling.
- **Finding 02 / 03 / 04** — Replace SHA-256 password hashing with
  Argon2id KDF + per-user salt, and verify passwords by attempting AEAD
  decryption rather than comparing stored hashes (`87e9fc34`).
- **Finding 05** — Clear sensitive key material from in-memory keyring
  state on wallet lock (`f2a0063a`).
- **Finding 06** — Restrict the `<meta name="gnoconnect:rpc">` meta-tag
  RPC injection path to a hardcoded dApp origin allowlist sourced from
  `chains.json` `gnoUrl` values, and remove the `UNADDED_NETWORK`
  fall-through in the content handler so an unknown chain ID can no
  longer slip through origin approval (`827c4e3f`).

The infrastructure commit `479b80da` introduces the `v018` storage model
and extends the storage migrator to support per-user salts required by
Finding 01.

### Trust model alignment

For Finding 06, the new origin allowlist mirrors the curated trust list
patterns used by Keplr (`EmbedChainInfos`) and MetaMask (`FEATURED_RPCS`):
the page is only allowed to influence the RPC endpoint when it is itself
served from a maintainer-approved origin. Adding a new gno.land
deployment requires updating
`packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts`.

The check is scoped to the meta-tag injection path
(`CommandHandler.createContentHandler`); standard injected-provider dApp
flows go through `CommandHandler.createHandler` and are unaffected. All
other `new GnoProvider(...)` call sites consume RPC URLs from the user's
saved `NetworkMetainfo` rather than page-supplied data.

## Findings deliberately not included in this PR

- **07 — Timing side-channel in password comparison.** Made obsolete by
  Findings 02/03/04 in this PR. `WalletService.equalsPassword` now
  verifies the password by attempting AEAD decryption, and the prior
  `encryptSha256Password(password) === storedPassword` site at
  `wallet.ts:253` is gone. The remaining string-equality uses in
  `validation-password.ts` either compare two client-side inputs
  (confirm-password, change-password equality) or are dead code, neither
  of which is exploitable. Both Keplr and MetaMask rely on KDF-gated
  decryption rather than constant-time hash comparison.
- **08 — Plaintext intermediate values during wallet serialization.**
  Out of scope for this PR. Eliminating the intermediate plaintext
  requires redesigning `Wallet.serialize` to encrypt per keyring instead
  of building a single JSON blob, which is a non-trivial refactor that
  should land separately. Finding 05's memzero on lock already shortens
  the residency window for the same key material.
- **09 — Background script does not validate message sender.** Worth
  fixing, but unrelated to the keyring/crypto cluster on this branch.
  Will be addressed in a focused follow-up that adds
  `sender.id`/`sender.frameId` checks with dedicated regression tests.
- **10 — 5-minute post-disconnect session timeout.** Marked Info by the
  reviewer; product/UX decision rather than a vulnerability. A
  user-configurable timeout is feature work, not a fix.
- **11 — `web_accessible_resources` exposes `resources/*` to all
  origins.** Info-level fingerprinting risk. Needs an audit of which
  resources are actually consumed by external pages versus only by the
  extension itself before tightening `matches`; deferred to avoid
  breaking dApp-facing assets.
- **12 — Content script runs on `<all_urls>`.** The reviewer explicitly
  notes this is expected wallet-extension behavior (same as MetaMask)
  and recommends no change. Acknowledged, no action required.
- **13 — Unused Web3Auth/Torus stub code and dependencies.** Product
  decision on whether Web3Auth integration is still planned. Removing
  the package belongs in a chore PR once that direction is confirmed; it
  is not a security regression caused by this branch.

## Test plan

- [ ] `npm test` in `packages/adena-extension`
- [ ] `npm test` in `packages/adena-module`
- [ ] Manual: existing wallet locked with the previous AES-CBC scheme
      successfully migrates to v018 on next unlock
- [ ] Manual: password change / login / wallet export / approve-login
      flows all succeed with the Argon2id-based verification
- [ ] Manual: lock the wallet and confirm sensitive material is cleared
      from keyring instances
- [ ] Manual: `https://gno.land`, `https://betanet.testnets.gno.land`,
      `https://staging.gno.land`, `https://test13.testnets.gno.land`
      gnoweb action triggers open the transaction modal as expected
- [ ] Manual: a page on an unlisted origin that declares
      `<meta name="gnoconnect:rpc" content="https://attacker.example/rpc">`
      does **not** open the transaction modal
- [ ] Manual: standard injected-provider dApp interactions
      (`window.adena.*`) continue to work without change

### DIFF
--- a/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
+++ b/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
@@ -1,0 +1,10 @@
+// dApp origin allowlist for gnoconnect meta-tag based RPC injection.
+// Pages served from these origins may declare their preferred RPC and chain ID
+// via <meta name="gnoconnect:rpc"> / <meta name="gnoconnect:chainid">.
+// Initial set is sourced from chains.json gnoUrl values.
+export const GNO_CONNECT_ALLOWED_ORIGINS: readonly string[] = [
+    'https://gno.land',
+    'https://betanet.testnets.gno.land',
+    'https://staging.gno.land',
+    'https://test13.testnets.gno.land',
+] as const;
--- a/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
+++ b/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
@@ -1,10 +1,17 @@
-// dApp origin allowlist for gnoconnect meta-tag based RPC injection.
-// Pages served from these origins may declare their preferred RPC and chain ID
-// via <meta name="gnoconnect:rpc"> / <meta name="gnoconnect:chainid">.
-// Initial set is sourced from chains.json gnoUrl values.
-export const GNO_CONNECT_ALLOWED_ORIGINS: readonly string[] = [
-    'https://gno.land',
-    'https://betanet.testnets.gno.land',
-    'https://staging.gno.land',
-    'https://test13.testnets.gno.land',
-] as const;
+import CHAIN_DATA from '@resources/chains/chains.json';
+
+// dApp origin allowlist for gnoconnect meta-tag RPC injection.
+// Derived from chains.json gnoUrl. Loopback origins (127.0.0.1, localhost)
+// are trusted only in development builds — in production any process
+// holding the local port (other dev servers, malicious postinstall, etc.)
+// could otherwise inject RPC endpoints into the wallet flow.
+const isDevelopmentBuild = process.env.NODE_ENV !== 'production';
+
+const isLoopbackOrigin = (url: string): boolean =>
+  url.startsWith('http://127.0.0.1') || url.startsWith('http://localhost');
+
+export const GNO_CONNECT_ALLOWED_ORIGINS: readonly string[] = CHAIN_DATA.map(
+  (chain) => chain.gnoUrl,
+)
+  .filter((url): url is string => Boolean(url))
+  .filter((url) => url.startsWith('https://') || (isDevelopmentBuild && isLoopbackOrigin(url)));
--- a/packages/adena-extension/src/common/storage/chrome-local-storage.ts
+++ b/packages/adena-extension/src/common/storage/chrome-local-storage.ts
@@ -22,7 +22,8 @@ type StorageKeyTypes =
   | 'ADD_ACCOUNT_GUIDE_CONFIRM_DATE'
   | 'ACCOUNT_GRC721_COLLECTIONS'
   | 'ACCOUNT_GRC721_PINNED_PACKAGES'
-  | 'KDF_SALT';
+  | 'KDF_SALT'
+  | 'AUTO_LOCK_TIMEOUT_MINUTES';
 
 // List of all available storage keys
 const StorageKeys: StorageKeyTypes[] = [
@@ -45,6 +46,7 @@ const StorageKeys: StorageKeyTypes[] = [
   'ACCOUNT_GRC721_COLLECTIONS',
   'ACCOUNT_GRC721_PINNED_PACKAGES',
   'KDF_SALT',
+  'AUTO_LOCK_TIMEOUT_MINUTES',
 ];
 
 // Function to validate if a given key is a valid storage key
--- a/packages/adena-extension/src/common/storage/chrome-local-storage.ts
+++ b/packages/adena-extension/src/common/storage/chrome-local-storage.ts
@@ -11,7 +11,6 @@ type StorageKeyTypes =
   | 'CURRENT_NETWORK_ID'
   | 'CURRENT_ATOMONE_NETWORK_ID'
   | 'SERIALIZED'
-  | 'ENCRYPTED_STORED_PASSWORD'
   | 'CURRENT_ACCOUNT_ID'
   | 'ACCOUNT_NAMES'
   | 'ESTABLISH_SITES'
@@ -22,7 +21,8 @@ type StorageKeyTypes =
   | 'WALLET_CREATION_GUIDE_CONFIRM_DATE'
   | 'ADD_ACCOUNT_GUIDE_CONFIRM_DATE'
   | 'ACCOUNT_GRC721_COLLECTIONS'
-  | 'ACCOUNT_GRC721_PINNED_PACKAGES';
+  | 'ACCOUNT_GRC721_PINNED_PACKAGES'
+  | 'KDF_SALT';
 
 // List of all available storage keys
 const StorageKeys: StorageKeyTypes[] = [
@@ -33,7 +33,6 @@ const StorageKeys: StorageKeyTypes[] = [
   'CURRENT_NETWORK_ID',
   'CURRENT_ATOMONE_NETWORK_ID',
   'SERIALIZED',
-  'ENCRYPTED_STORED_PASSWORD',
   'CURRENT_ACCOUNT_ID',
   'ACCOUNT_NAMES',
   'ESTABLISH_SITES',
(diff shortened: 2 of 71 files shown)


dongwon8247 (approved):

packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
```diff
@@ -0,0 +1,10 @@
+// dApp origin allowlist for gnoconnect meta-tag based RPC injection.
+// Pages served from these origins may declare their preferred RPC and chain ID
+// via <meta name="gnoconnect:rpc"> / <meta name="gnoconnect:chainid">.
+// Initial set is sourced from chains.json gnoUrl values.
+export const GNO_CONNECT_ALLOWED_ORIGINS: readonly string[] = [
+    'https://gno.land',
+    'https://betanet.testnets.gno.land',
+    'https://staging.gno.land',
+    'https://test13.testnets.gno.land',
+] as const;
```
jinoosss: Should it be managed differently from chains.json?
https://github.com/onbloc/adena-wallet/blob/f1c6bf1bb32df9c0513336da75650c1982700178/packages/adena-extension/src/resources/chains/chains.json

packages/adena-extension/src/common/storage/chrome-local-storage.ts
```diff
@@ -96,6 +96,11 @@ export class ChromeLocalStorage implements Storage {
   public updatePassword = async (password: string): Promise<StorageModelLatest | null> => {
     const current = await this.migrator.getCurrent();
     this.current = await this.migrator.migrate(current, password);
+    if (this.current) {
+      await this.storage.set({
+        [ChromeLocalStorage.StorageKey]: JSON.stringify(this.current),
+      });
+    }
```
jinoosss, line 105: This logic can update the state during a simple password check.
I expect the `this.migrator.migrate` function to fail, but is that fine?
https://github.com/onbloc/adena-wallet/blob/8ee88e747a388b48f485d4a9513e17c1d7059a70/packages/adena-extension/src/services/wallet/wallet.ts#L276-L279

packages/adena-extension/src/hooks/web/connect-ledger/use-select-account-screen.ts
```diff
@@ -159,9 +159,10 @@ const useSelectAccountScreen = (): useSelectAccountScreenReturn => {
         keyrings: [keyring.toData()],
         currentAccountId: resultSavedAccounts[0]?.id,
       });
+      const serializedWallet = newWallet.toJSON();
       navigate(RoutePath.WebCreatePassword, {
         state: {
-          serializedWallet: await newWallet.serialize(''),
+          serializedWallet,
           stepLength: 5,
```
jinoosss: We are handling wallet JSON data as a parameter when navigating between pages.
This data is not being reset after use, and it appears that it could be recorded in a memory capture.

packages/adena-extension/src/migrates/migrations/v018/storage-migration-v018.ts
```diff
@@ -0,0 +1,170 @@
...
+      plaintext = await decryptAES(serialized, oldPassword);
+    } catch {
+      throw new Error('V018 migration failed: cannot decrypt SERIALIZED with given password');
+    }
+    if (!plaintext || plaintext.trim() === '') {
+      throw new Error('V018 migration failed: cannot decrypt SERIALIZED with given password');
+    }
+
+    const encrypted = await encryptXChacha20(plaintext, newPassword, salt);
+    return JSON.stringify(encrypted);
```
jinoosss, line 138: Please add the following line: `plaintext = '';`.

jinoosss (commented):
LGTM ! The encryption method has been changed, allowing you to manage your data securely.

packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
HeesungB: Good point — switched in 547921a. The list is now derived from chains.json gnoUrl values, with https:// always trusted and loopback origins (127.0.0.1, localhost) included only in development builds. Production users no longer have a local-port path to RPC injection, while dev keeps the local node usable.

packages/adena-extension/src/common/storage/chrome-local-storage.ts
HeesungB: You're right that state can change during verification. It's intentional today (the call triggers the v017→v018 re-encryption on first unlock), and storage data itself stays safe — `migrate` throws on wrong password and the `if (this.current)` guard skips `storage.set`, so on-disk data is not mutated.

Two things still worth addressing, though:
1. Naming — `updatePassword` reads as a simple setter but actually performs a migration; renaming it (e.g. `migrateWithPassword`) would surface the cost to callers.
2. Backup churn — `migrator.migrate`'s catch path calls `backup(current)` on every failure, including wrong-password decryption errors, so each typo accumulates an `ADENA_DATA_<timestamp>` entry. The backup should only fire on real schema corruption.

Both are out of scope for this PR; tracking as a follow-up.

packages/adena-extension/src/hooks/web/connect-ledger/use-select-account-screen.ts
HeesungB: Valid concern. The same `toJSON()` pattern actually spans five entry hooks (create / import-mnemonic / import-privateKey / google-login / airgap / ledger), and all of them route into `WebCreatePassword`. Added `window.history.replaceState({}, '')` in the receiver right after `saveWallet`, so the keyring JSON is dropped from the back-stack once it's been committed to encrypted storage (b14f449d).

This shortens the residency window but doesn't eliminate it during password entry. A stronger fix (e.g. an ephemeral in-memory store) will land in a separate PR.

packages/adena-extension/src/migrates/migrations/v018/storage-migration-v018.ts
HeesungB: Done in 02586901. Applied the same line in `migrateAddressBook` (line 167) too, since both functions hold a transient plaintext after `encryptXChacha20`. Effect is GC-eligibility rather than true zeroization (V8 strings are immutable), but it matches the `rawSeeds = ''` pattern used elsewhere.

packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
HeesungB: Followup landed in this PR (67a8d8b2). Replaced the `history.state` handoff with a module-scope `pendingWalletStore` — entry hooks call `set(wallet)` and `navigate(WebCreatePassword)` with no state, the receiver `consume()`s at save time, and `_saveWalletByPassword` wraps `saveWallet` in `try/finally { wallet.destroy() }` so `sodium.memzero` wipes the Uint8Array seed/entropy on both success and failure. Mount-time guard redirects refresh/direct-entry, unmount cleanup destroys any pending wallet. `toJSON()`/`fromJSON()` and `history.replaceState` are gone from the path.

packages/adena-extension/src/common/storage/chrome-local-storage.ts
jinoosss: Thank you for your clear answer.

jinoosss (dismissed):

jinoosss (approved):

PR #837, opened 2026-04-30
[ADN-745] feat: customizable auto-lock timer

HeesungB:
## Summary
- Replace the hard-coded 5-minute password expiry with a `chrome.alarms` one-shot idle timer that survives SW restarts and laptop sleep.
- Add **Settings → Security & Privacy → Auto-Lock Timer** page (0 ~ 1440 min, digit-only input, Figma error copy). 0 disables auto-lock.
- Background broadcasts `AUTO_LOCK_TRIGGERED` on lock; UI invalidates the cached `['wallet/locked']` query so Login renders immediately.

## Test plan
- [ ] 1-min timer → popup idle for 1 min auto-routes to Login
- [ ] SW terminate / laptop sleep → wallet still locks on time
- [ ] 0 → never auto-locks; red "disabled" notice shown
- [ ] `-`, `.`, `e`, arrow keys, wheel, paste `1.5` → all rejected
- [ ] `1441` → range error; Save disabled
- [ ] Existing user (no key) → defaults to 5
- [ ] `yarn test:ci` and `yarn build` pass

## Notes
Based on `fix/ADN-747`. Set the PR base to `fix/ADN-747` so only the four ADN-745 commits show up; rebase onto `main` after that branch merges.

### DIFF
--- a/packages/adena-extension/package.json
+++ b/packages/adena-extension/package.json
@@ -93,6 +93,7 @@
     "qrcode.react": "3.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-idle-timer": "5.7.2",
     "react-router-dom": "6.22.1",
     "recoil": "0.7.7",
     "styled-components": "6.1.14",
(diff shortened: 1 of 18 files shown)


dongwon8247 (approved):

jinoosss (approved):
LGTM !

PR #838, opened 2026-04-30
fix: auto-detect chain from address prefix in address book (ADN-765)

HeesungB:
## Summary

- Auto-detect chain from a saved/typed address prefix so the Add/Edit Address screen no longer mislabels atomone (`atone1...`) entries as Gno.land.
- Derive `addressMaxLength` from the longest registered bech32 prefix so pasting a longer-prefix address (e.g. `atone1...` while the form still shows gno) is no longer truncated by the textarea before chain auto-sync runs.
- Restructure the Edit Address layout: move `Remove Address` out of absolute positioning, pin Cancel/Save to the bottom, and make the rest of the form scroll so longer 2-line atomone addresses no longer overlap the action row.

## Background

`AddressBookItem` stores the bare address with no `chainGroup` field, so the chain has to be inferred from the bech32 prefix at display time. The list view (`address-book/index.tsx`) already did this with a local `inferChainGroup` helper; the Add/Edit form did not, defaulting to `gno` and disabling the dropdown in Edit mode. That stale value also constrained `addressMaxLength` (40 vs. 44) and was passed into `validateAlreadyAddressByAccounts(..., chain.bech32Prefix)`, silently misclassifying duplicate-account checks for atomone.

This change centralizes the heuristic in `common/utils/address-chain.ts`, seeds `chainGroup` from `params.curr?.address` on mount, and re-syncs on every address change. The dropdown remains user-clickable in Add mode (typing/pasting still wins) and disabled in Edit mode (existing behavior).

## Changes

- **New**: `packages/adena-extension/src/common/utils/address-chain.ts` — shared `inferChainGroup` helper.
- **Modified**: `pages/popup/certify/add-address/index.tsx`
  - Seed `chainGroup` from saved address; `useEffect` re-syncs on address change.
  - `addressMaxLength` derived from `chainRegistry.listChains()` (longest prefix), preventing paste truncation across chains.
  - `AddressInput`: `min-height: 70px; height: auto; word-break: break-all;` so 44-char atomone addresses wrap cleanly on two lines.
  - `RemoveAddressBtn`: out of absolute positioning, into normal flow with margins.
  - New `ScrollableContent` wrapper: middle of the form scrolls, Cancel/Save stays pinned at the bottom.
- **Modified**: `pages/popup/certify/address-book/index.tsx` — drop the duplicated local `inferChainGroup`, import from the shared util.

## Test plan

- [ ] Address Book → Add Address → paste an `atone1...` address → Chain dropdown switches to AtomOne, full 44 chars accepted (no truncation).
- [ ] Address Book → Add Address → paste a `g1...` address → Chain stays Gno.land, full 40 chars accepted.
- [ ] Save an atomone entry, re-open via Edit → Chain shows AtomOne, address renders on two lines, `Remove Address` no longer overlaps with the address text.
- [ ] In Edit Address, Cancel/Save remain pinned at the bottom; scroll the middle area to confirm the form content scrolls independently.
- [ ] Address Book list view: existing chain-icon inference for both gno and atomone entries still works (regression check).
- [ ] Save flow with both gno and atomone entries: no false `ALREADY_ADDRESS` from `validateAlreadyAddressByAccounts` (which now receives the right prefix).

### DIFF
--- a/packages/adena-extension/src/common/utils/address-chain.ts
+++ b/packages/adena-extension/src/common/utils/address-chain.ts
@@ -1,7 +1,15 @@
+import type { ChainRegistry } from 'adena-module';
+
 // Returns the chainGroup that owns the given bech32 address prefix.
+// Sorts registered chains by prefix length (longest first) so a chain
+// with a longer prefix wins over one whose prefix is a substring of it.
 // Falls back to 'gno' for unknown or partial input so the form keeps a
 // usable default while the user is still typing.
-export function inferChainGroup(address: string): string {
-  if (address.startsWith('atone')) return 'atomone';
-  return 'gno';
+export function inferChainGroup(address: string, chainRegistry: ChainRegistry): string {
+  const matched = chainRegistry
+    .listChains()
+    .slice()
+    .sort((a, b) => b.bech32Prefix.length - a.bech32Prefix.length)
+    .find((chain) => address.startsWith(chain.bech32Prefix + '1'));
+  return matched?.chainGroup ?? 'gno';
 }
--- a/packages/adena-extension/src/common/utils/address-chain.ts
+++ b/packages/adena-extension/src/common/utils/address-chain.ts
@@ -1,0 +1,7 @@
+// Returns the chainGroup that owns the given bech32 address prefix.
+// Falls back to 'gno' for unknown or partial input so the form keeps a
+// usable default while the user is still typing.
+export function inferChainGroup(address: string): string {
+  if (address.startsWith('atone')) return 'atomone';
+  return 'gno';
+}
--- a/packages/adena-extension/src/common/utils/address-chain.spec.ts
+++ b/packages/adena-extension/src/common/utils/address-chain.spec.ts
@@ -1,0 +1,29 @@
+import { createChainRegistry } from 'adena-module';
+import { inferChainGroup } from './address-chain';
+
+describe('inferChainGroup', () => {
+  const registry = createChainRegistry();
+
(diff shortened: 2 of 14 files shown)


dongwon8247 (approved):

packages/adena-extension/src/common/utils/address-chain.ts
```diff
@@ -0,0 +1,7 @@
+// Returns the chainGroup that owns the given bech32 address prefix.
+// Falls back to 'gno' for unknown or partial input so the form keeps a
+// usable default while the user is still typing.
+export function inferChainGroup(address: string): string {
+  if (address.startsWith('atone')) return 'atomone';
+  return 'gno';
+}
```
jinoosss: It would be a good idea to use the data from `chains.json` and `atomone-chains.json`.

jinoosss (commented):
LGTM !

packages/adena-extension/src/common/utils/address-chain.ts
HeesungB: Done in 1a7bee02. Switched to a ChainRegistry-driven lookup so chains.json / atomone-chains.json stay the single source of truth — `createChainRegistry()` already registers both. The function now sorts `listChains()` by `bech32Prefix` length (longest first) and matches on `prefix + '1'`, which guards against future cases where one chain's prefix is a substring of another's. The three call sites (address-book list, add-address form, address-book input hook) all receive `chainRegistry` via the existing `useAdenaContext()` channel. Added a small spec covering gno / atomone / unknown / empty / partial inputs.

jinoosss (approved):

PR #844, opened 2026-05-13
fix: default network edit/delete robustness (ADN-776)

HeesungB:
## Summary

- Preserve user edits to default networks (rpc/rest URLs, deleted flag) across reload, and wire the **Reset to Default** button on the Edit Network screen to actually restore the shipped `chains.json` values.
- Refresh `chainRegistry` and invalidate the React Query balance caches when the active gno/atomone network is edited so the new RPC / Chain ID / REST URL takes effect immediately instead of after the next refetch interval. dApp `approveEstablish` health checks also pick up the new endpoint.
- When deleting the active network, fall back to the first non-deleted entry matching the current testnet mode (was hardcoded beta-mainnet for gno, hardcoded `isMainnet` for atomone). The new current network id is also persisted so the fallback survives a reload.

## Test plan

- [ ] Edit active gno default network's RPC URL -> Save -> balance refetches from the new RPC without waiting for the refetch interval
- [ ] Edit active gno default network's Chain ID -> dApp Approve Establish health check uses the new chainId/RPC
- [ ] Edit active atomone network's RPC/REST URL -> cosmos balance refetches from the new endpoints immediately
- [ ] Edit a default network -> reload the extension -> user edits and `deleted` flag survive
- [ ] Edit a default network -> click **Reset to Default** -> storage is restored to the `chains.json` factory values
- [ ] testnet mode ON: activate a custom gno testnet, delete it -> wallet falls back to the first non-deleted gno testnet (not beta-mainnet)
- [ ] mainnet mode: same path -> wallet falls back to the first gno mainnet
- [ ] Repeat the two cases above for atomone (testnet mode delete -> atomone testnet, not atomone mainnet)
- [ ] After delete fallback, reload the extension -> the fallback selection persists (current network id was saved)
- [ ] Regression: switching to an unedited network still works for both chain groups
- [ ] Regression: adding a brand-new custom network (gno + atomone) still works and shows up in the list

### DIFF
--- a/packages/adena-extension/src/hooks/use-network.ts
+++ b/packages/adena-extension/src/hooks/use-network.ts
@@ -51,6 +51,7 @@ interface NetworkResponse {
   changeNetwork: (networkId: string) => Promise<boolean>;
   changeNetworkMode: (mode: NetworkMode) => Promise<void>;
   updateNetwork: (network: NetworkMetainfo | AtomoneNetworkMetainfo) => Promise<boolean>;
+  resetNetworkToDefault: (chainGroup: ChainGroup, networkId: string) => Promise<boolean>;
   deleteNetwork: (chainGroup: ChainGroup, networkId: string) => Promise<boolean>;
   setModified: (modified: boolean) => void;
 }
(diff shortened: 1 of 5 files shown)


packages/adena-extension/src/hooks/use-network.ts
```diff
@@ -358,11 +397,17 @@ export const useNetwork = (): NetworkResponse => {
...
+          const isMainnet = mode === 'mainnet';
           const fallback =
-            changedNetworks.find((current) => !current.deleted && current.isMainnet) ?? null;
+            changedNetworks.find(
+              (current) => !current.deleted && current.isMainnet === isMainnet,
+            ) ??
+            changedNetworks.find((current) => !current.deleted) ??
+            null;
           setCurrentAtomoneNetwork(fallback);
           if (fallback) {
```
jinoosss, line 408: Why is the fallback handling different?
https://github.com/onbloc/adena-wallet/pull/844/changes#diff-f09cd4f95dc28bffd311da12562e566b7d55a4b179902b94dda174472ddac0c4R438-R444

HeesungB: The asymmetry comes from `changeNetworkOfProvider` being gno-only. It wraps three things: rebuild the `GnoProvider` RPC client, update Recoil `currentNetwork`, and dispatch the `changedNetwork` event to connected dApps. atomone has no equivalent provider instance and isn't part of the dApp event channel, so it just calls `setCurrentAtomoneNetwork` directly. This follows the same gno-vs-atomone split already used in `changeNetwork`, `changeNetworkMode`, and `syncModeFromTarget` in the same file - it isn't a new divergence introduced by this PR.

jinoosss (approved):

PR #849, opened 2026-05-19
feat(ADN-780): default fresh installs to Testnet Mode with test-13

HeesungB:
## Summary

Fresh installs now boot into Testnet Mode with `test-13` selected, instead of Beta Mainnet/Staging.

Existing users are unaffected: when `NETWORK_MODE` is unset, the active mode is derived from the stored network's `main` flag, so a user who has only ever used Beta Mainnet stays on Beta Mainnet after the update.

### What changed

- **chains.json** - `test-13` now precedes `staging` so the testnet default list naturally falls on `test-13`.
- **`network-default` helper (new)** - `pickDefaultByMode` (prefers canonical ids `test-13` / `gnoland1`, falls back to generic + first non-deleted), `resolveNetworkMode` (stored mode -> derive from stored network -> testnet for fresh installs), `normalizeStoredId` (handles `''`, `'undefined'`, `'null'` sentinels).
- **`WalletProvider` bootstrap** - load gno networks, load AtomOne networks, resolve mode using both stored values, then pick the current network per chain group sequentially. Removes the race between `initNetworkMetainfos` and the old `initAtomoneNetworkChain`.
- **`useNetwork`** - `changeNetworkMode` now uses `pickDefaultByMode` so toggling Testnet always lands on `test-13`. The hook's `currentNetwork` fallback is mode-aware to avoid a gnoland1 flash before the provider hydrates.
- **`currentAtomoneNetwork` atom default -> `null`** so the first render does not flash a mainnet AtomOne entry on fresh-install testnet users.
- **`getCurrentNetworkId` repository method** - normalizes `'undefined'` / `'null'` / empty values, mirroring the AtomOne variant. `StorageManager.get` stringifies missing values, which previously slipped through truthy checks.

### Mode resolution policy

| `NETWORK_MODE` | `CURRENT_NETWORK_ID` | Resolved mode |
|---|---|---|
| `'mainnet'` / `'testnet'` | any | the stored mode |
| unset | known id | derive from stored network's `main` flag |
| unset | unset | `testnet` (fresh install) |

## Test plan

Automated (passing locally):
- [x] `pickDefaultByMode`, `resolveNetworkMode`, `normalizeStoredId` unit tests (`network-default.spec.ts`)
- [x] `getCurrentNetworkId` normalization tests (`chain.spec.ts`)
- [x] Full extension test suite: `npx jest` -> 138 suites / 382 tests pass
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint` on changed files clean

Manual:
- [ ] Fresh install: clear storage -> create wallet -> verify Testnet toggle ON, current network `Testnet 13`
- [ ] Toggle Testnet OFF -> `gnoland1` (Beta Mainnet); toggle ON -> back to `test-13`
- [ ] Existing user simulation: storage has `CURRENT_NETWORK_ID=gnoland1`, no `NETWORK_MODE` -> verify toggle OFF, current = gnoland1
- [ ] Existing user simulation: storage has `CURRENT_NETWORK_ID=staging`, no `NETWORK_MODE` -> verify toggle ON, current = staging
- [ ] Explicit selection preserved: with both `NETWORK_MODE` and `CURRENT_NETWORK_ID` set, no override on reload

### DIFF
--- a/packages/adena-extension/src/hooks/use-network.ts
+++ b/packages/adena-extension/src/hooks/use-network.ts
@@ -2,6 +2,7 @@ import { useCallback, useMemo } from 'react';
 import { useRecoilState, useRecoilValue } from 'recoil';
 
 import { fetchHealth } from '@common/utils/fetch-utils';
+import { pickDefaultByMode } from '@common/utils/network-default';
 import { EventMessage } from '@inject/message';
 import { useAdenaContext, useWalletContext } from './use-context';
 import { useEvent } from './use-event';
@@ -57,7 +58,10 @@ interface NetworkResponse {
   setModified: (modified: boolean) => void;
 }
 
-const DEFAULT_NETWORK: NetworkMetainfo = CHAIN_DATA[0];
+const DEFAULT_TESTNET_NETWORK: NetworkMetainfo =
+  CHAIN_DATA.find((network) => network.id === 'test-13') ?? CHAIN_DATA[0];
+const DEFAULT_MAINNET_NETWORK: NetworkMetainfo =
+  CHAIN_DATA.find((network) => network.id === 'gnoland1') ?? CHAIN_DATA[0];
 
 function isAtomoneNetwork(
   network: NetworkMetainfo | AtomoneNetworkMetainfo,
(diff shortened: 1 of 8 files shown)


packages/adena-extension/src/hooks/use-network.ts
```diff
@@ -57,7 +58,10 @@ interface NetworkResponse {
   setModified: (modified: boolean) => void;
 }
 
-const DEFAULT_NETWORK: NetworkMetainfo = CHAIN_DATA[0];
+const DEFAULT_TESTNET_NETWORK: NetworkMetainfo =
+  CHAIN_DATA.find((network) => network.id === 'test-13') ?? CHAIN_DATA[0];
+const DEFAULT_MAINNET_NETWORK: NetworkMetainfo =
+  CHAIN_DATA.find((network) => network.id === 'gnoland1') ?? CHAIN_DATA[0];
```
jinoosss, line 65: It would be great if this could be improved in the next update.

jinoosss (approved):

PR #853, opened 2026-05-29
feat(ADN-769): add session account support

HeesungB:
### What type of PR is this?

feature

### What this PR does:

Adds **session account** support to Adena Wallet (ADN-769).

A session account is a restricted, time-limited sub-account derived from a master
account. It holds its own key pair (not the master key) and can only sign within
limits agreed in advance: an allowlist of contract paths, a spending cap, and an
expiration. On-chain, its transactions carry a `LocalTxSignature` with a
`session_addr` field so the chain knows which session signed.

This lets a user delegate a narrow, revocable signing capability to a dApp/session
without exposing the master account.

#### Key changes

- **Core (adena-module)** — session tx wire encoding (`LocalTxSignature`, session
  message types) and a `SessionKeyring` / `SessionAccount` model.
- **Storage** — `v020` migration and a dedicated `SessionRepository`.
- **Signing guard** — enforces allowlisted paths and spend limits before signing.
- **Lifecycle** — session import (preview + commit), master to session conversion,
  and revoke flows.
- **dApp safety** — injected provider methods (cosmos / multisig / network / wallet)
  reject session accounts, so dApps cannot drive a session account directly.
- **UI** — session detail card, header session-overview popover, manage-sessions and
  revoke pages, and the session-add / advanced-setup web screens.

#### How to review

Commits are split bottom-up by architecture layer so they can be read in order
(each builds on the previous):

| # | Commit | Layer |
|---|--------|-------|
| 1 | session tx encoding and message types (adena-module) | module: proto / utils |
| 2 | session keyring and account model (adena-module) | module: keyring / account |
| 3 | session gno provider queries and common utils | ext: provider / common |
| 4 | session storage migration and repository | ext: storage |
| 5 | session signing guard and spend services | ext: services |
| 6 | session import and conversion wallet services | ext: services |
| 7 | session react hooks | ext: hooks |
| 8 | guard injected provider methods against session accounts | ext: inject |
| 9 | session UI components and assets | ext: components |
| 10 | session management and import pages | ext: pages |

#### Notes

- New storage migration `v020`; please verify upgrade from a populated pre-v020 profile.
- Session accounts are restricted to supported chains only.
- This branch also carries one preceding commit from ADN-780
  (`feat(ADN-780): default fresh installs to Testnet Mode with test-13`) as its base,
  since ADN-769 was branched off ADN-780.


packages/adena-extension/src/pages/popup/wallet/approve-sign-transaction/index.tsx
```diff
@@ -282,6 +366,50 @@ const ApproveSignTransactionContainer: React.FC = () => {
...
     }
 
+    // Session race-condition re-check: between popup load and confirm
+    // click, the session may have expired or another tab may have pushed
+    // spendUsed over the limit. Re-evaluate the guard right before signing
+    // and abort if it no longer passes.
+    if (isSessionAccount(currentAccount) && currentNetwork) {
+      try {
+        const sessionAddr = await currentAccount.getAddress(chain.bech32Prefix);
+        const metadata = await sessionRepository.get(sessionAddr);
```
junghoon-vans: Critical: this pre-sign recheck still reads cached `SESSIONS` metadata, but `spendUsed`, `spendReset`, and `status` are chain-authoritative according to `SessionRepository`. If another tab/device spends the session or revokes it, this guard can approve stale state and still sign. Please fetch the current session from chain (`gnoProvider.getSession(...)`), sync/flatten it, and evaluate the guard against fresh chain state immediately before signing.

packages/adena-extension/src/services/transaction/session-spend.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (message.value.from_address !== signerAddress) return [];
+      return safeParse(message.value.amount);
+    }
+    case '/vm.m_call':
+    case '/vm.m_run': {
+      // vm caller carries gas fee separately; send is the coin outflow
+      // attributed to the signer. caller != signer means a different signer
+      // is paying — we still treat the send as the caller's outflow because
+      // the gno ante computes SpendForSigner on the signer position.
+      return safeParse(message.value.send);
```
junghoon-vans: Critical: the session spend-limit projection for VM messages only counts `send` and ignores `max_deposit`, even though `MsgCall`/`MsgRun` encode `max_deposit` into the signed transaction and the approval UI treats it as consumed value. A session can pass the wallet spend-limit check with `send=''` and a large `max_deposit`. Please include `max_deposit` in the projected spend or reject VM messages with non-empty `max_deposit` for session accounts.

junghoon-vans (changes requested):
Critical blockers found in session-account signing enforcement. Please address the inline comments before merge.

packages/adena-extension/src/inject/message/message-handler.ts
```diff
@@ -163,37 +166,54 @@ export class MessageHandler {
         });
         break;
       case 'CREATE_MULTISIG_ACCOUNT':
+        if (await HandlerMethod.rejectSessionAccountUnsupported(core, message, sendResponse)) {
```
junghoon-vans, line 169: Critical: this PR adds SessionAccount rejection to adjacent injected methods, but `SIGN_AMINO` just above still routes directly to `signAmino` without rejecting sessions or running `evaluateSessionSigningGuard`. That path later calls `createSignature(currentAccount, document)`, so a dApp can obtain a session-key signature without allowPaths, spendLimit, expiry/status, chain, or unsupported-message checks. Please either reject SessionAccount for `SIGN_AMINO` too or apply the same fail-closed guard immediately before signing.

packages/adena-extension/src/services/transaction/session-spend.ts
jinoosss: Referrence:
https://github.com/gnolang/gno/blob/f3d5a5d13c09e19fadcb836e8473eef3177ed429/gno.land/adr/adr-001-session-subaccounts.md#spend-limit-enforcement

jinoosss: 952dfbcb740a5820f44eece8101f622fd337f17a

packages/adena-extension/src/inject/message/message-handler.ts
jinoosss: da50df3dcee7d9dde7aed8f9b28d2ae5182b9dbb

packages/adena-extension/src/pages/popup/wallet/approve-sign-transaction/index.tsx
jinoosss: 1c3d07729c544c2d98373e1f983a6f216f7d9539

jinoosss (dismissed):

junghoon-vans (dismissed):

jinoosss (approved):

PR #866, opened 2026-07-08
feat: handle revoked session accounts and surface the master address consistently

jinoosss:
## Summary

1. **Revoked sessions are now a first-class state.** A session revoked on chain used to be silently converted into a regular account. It is now kept as a `SessionAccount`, flagged `REVOKED`, and the UI guides the user to recover their key and remove the account.
2. **A session account always shows its master address.** Every user-facing surface (header, popover, deposit, side menu, accounts page) now resolves to the master address instead of the session key's own address, which can never receive funds.

Also removes the rollout allowlist in favor of a denylist, so sessions are enabled everywhere except `gnoland1`.

## Revoked session handling

The chain deletes the session record on `MsgRevokeSession`, so "not found on chain" is the revoke signal.

- **Detection** — `useSessionRevocationWatcher` is mounted once in the popup header (which renders on every wallet screen) and polls the chain every 5s for the *current* session only. Two false positives are guarded against before flagging: a session created moments ago may not be queryable yet (creation grace), and a single missing read may be a transient node hiccup (recheck). `REVOKED` is terminal, so the poll stops once it is set.
- **Conversion removed** — `session-account-conversion.ts` and `useConvertSessionAccounts` are deleted. `useVisibleAccounts` collapses to a pure `chainId` filter; it previously fired an RPC sweep from each of its mount sites. The remaining conversion call sites (balance input, transfer summary, approve sign) now write `setStatus('REVOKED')`.
- **Balance** — an `ACTIVE` session spends the master's funds, so funding surfaces resolve to the master. A revoked session can no longer do that, so `currentFundingAddress` falls back to the session key's own address — the balance shown then belongs to the key the user can still export.
- **Main screen** — content is faded to 50% and inerted. The rule targets the children of `<main>`, not `<main>` itself, so the scroll container stays interactive and the header/footer (siblings) remain usable.
- **Session Overview popover** — allowed paths, spend limit and expiry describe a delegation the chain no longer honours, so the overview card is replaced by the `REVOKED` chip and two recovery actions per the design:
  - *remove this account* → Remove Account page
  - *export your key* → stages the private-key export and opens the security tab

### Bug found along the way

`removeAccount` picked the next current account with `clone.accounts[clone.accounts.length - 1]`, which is `undefined` once the last account is removed — the Remove button crashed on `.id`. Nothing reached this path before, but a revoked session account imported on its own is exactly that case, and the popover now routes users there. Removing the only account resets the wallet instead.

## Master address consistency

A session address holds no balance and can never receive a deposit, so exposing it as a copyable address invites stranded deposits.

- `useAccountChainAddresses` returns a single Gno/master entry for session accounts. The AtomOne row is hidden entirely — a session has no key material to derive a non-Gno address. This also narrows the "Deposit by network" list to the Gno row.
- The header copy icon carried `aria-label='Copy address'` but had **no click handler**; it now copies the funding address.
- The Deposit screen already encoded the master address in the QR but labelled it with the session's nickname (`Session 1 (g1…)`). It now reads `Master` with an info tooltip.
- The side menu and accounts page rendered — and copied, and opened on GnoScan — the session address, while the balance beside it was the master's. Both now resolve through `getWalletFundingAddress`.


packages/adena-extension/src/hooks/use-account-list-infos.ts
```diff
@@ -31,7 +32,10 @@ const buildAccountListInfos = async (
     accounts.map(async (account) => ({
       accountId: account.id,
       name: formatNickname(accountNames[account.id] || account.name, 10),
-      address: await account.getAddress(addressPrefix),
+      // A SessionAccount row shows the master address — the one that holds the
+      // balance rendered next to it, and the only address worth copying or
+      // opening on GnoScan. Its own session address stays internal.
+      address: await getWalletFundingAddress(account, addressPrefix),
```
junghoon-vans, line 38: This now renders the master address for every SessionAccount row, which matches active sessions, but the balance rendered next to it still comes from `accountNativeBalanceMap`. That map unconditionally fetches the master balance for all SessionAccounts in `use-token-balance.ts`, while `useCurrentAccount` switches revoked sessions to the session-key balance. As a result, a revoked session can show the session-key balance on the main screen but the master balance in the account list/side menu. Please make the account-list balance path use the same revoked-session exception, or share the same balance-address helper/policy.

packages/adena-extension/src/common/constants/chain.constant.ts
```diff
@@ -2,7 +2,9 @@
...
 
-// Chain id on which the account-session feature is currently enabled during
-// rollout. Kept in one place so it is easy to widen (or move into chains.json
-// as a per-chain capability flag) when the feature ships to more networks.
-export const SESSION_SUPPORTED_CHAIN_ID = 'test-13';
+// Chains where the account-session feature must stay off. Everything else is
+// treated as supported, so a newly added (or custom) gno chain opts in by
+// default. Single source of truth: `isSessionSupportedNetwork` and
+// `isSessionSupportedChainId` in @common/utils/account-session are the only
+// readers.
```
junghoon-vans, line 9: Since this changes session support from a single allowlisted chain to supported-by-default, the existing network-method guard gap becomes wider: `SWITCH_NETWORK` rejects SessionAccounts in both the background handler and the approve popup, but `ADD_NETWORK` still has no `rejectSessionAccountUnsupported` check and `approve-adding-network` does not re-check after unlock. Please align `ADD_NETWORK` with `SWITCH_NETWORK` so dApps cannot drive network mutations while a SessionAccount is selected.

packages/adena-extension/src/hooks/wallet/use-session-revocation-watcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        gnoProvider
+          .getSession(masterAddress, sessionAddr)
+          .then((res) => !!res)
+          .catch(() => true), // treat a failed recheck as "still there"
+      );
+      if (!revoked) {
+        return 'unknown';
+      }
+
+      await sessionRepository.setStatus(sessionAddr, 'REVOKED').catch(() => undefined);
```
junghoon-vans: This swallows a failed `setStatus('REVOKED')`, but the query still returns terminal `revoked`, which disables both polling and focus refetch below. If the storage write fails transiently, the UI can stop retrying while the durable `SESSIONS` row remains `ACTIVE`. Please let the write failure keep the result non-terminal, or otherwise retry until the revoked status is persisted.

junghoon-vans (changes requested):
I left three inline comments. The address consistency changes are good overall, but these cases should be fixed before merge.

packages/adena-extension/src/common/constants/chain.constant.ts
jinoosss: 3ea26ba557ba439228eda5f78ad432dec5e73b88

packages/adena-extension/src/hooks/use-account-list-infos.ts
jinoosss: 53dc7a7b11af78b9f8b46470cbb87f4f2c29051f

packages/adena-extension/src/hooks/wallet/use-session-revocation-watcher.ts
jinoosss: a355c63918acce0acab49d17d245a46f86a1d1f4

junghoon-vans (dismissed):
Re-reviewed the follow-up commits. The three issues I raised are addressed: revoked-session balance policy is shared by the main screen and account list, ADD_NETWORK now has SessionAccount guards at the same layers as SWITCH_NETWORK, and the revocation watcher keeps polling if persisting REVOKED fails. Focused tests and CI are passing.

junghoon-vans (approved):
Re-reviewed the latest follow-up commits. I did not find any new blocking issues; the previous session-account findings remain addressed. Focused tests and targeted lint passed locally. The current GitHub check failures appear to be runner acquisition issues, not code failures.

PR #868, opened 2026-07-09
fix: support gnoconnect tx links on the Local network with a runtime-gated loopback trust

jinoosss:
## Summary

gnoconnect **tx links** (clicking a `$help&func=...` link on gnoweb to open the
signing popup) did not work against a **Local** gno node. This PR makes them work
while keeping the origin-trust hardening introduced in the security review (#836).

Two independent bugs blocked the Local flow, plus a security-conscious redesign of
how loopback origins are trusted.

### 1. Loopback origin was blocked by the gnoconnect origin gate

`command-handler` gates the tx-link flow on `isAllowedGnoConnectOrigin(origin)`.
The allowlist trusted loopback origins only in development builds, so a
production build silently dropped the flow for `http://127.0.0.1:8888` (the
built-in Local network), even though `dev` ships as a default network.

**Fix:** the allowlist is now derived purely from `chains.json`, and loopback
trust no longer depends on the build mode (see the runtime gate below for how it
is kept safe).

### 2. Local RPC endpoint reached the RPC client without a protocol

gnoweb declares the RPC via the `gnoconnect:rpc` meta tag. Local gno nodes
commonly advertise it as `tcp://127.0.0.1:26657` (or without any scheme). The old
`getUrlPathWithoutProtocol` stripped non-http schemes down to a bare host, which
then flowed into `GnoProvider` / `networkInfo.rpcUrl`. Downstream RPC clients
(`@gnolang/tm2-rpc` `HttpClient`, `fetch`, `axios`) reject an endpoint without an
`http(s)://` protocol, throwing:

> `Endpoint URL is missing a protocol. Expected 'https://' or 'http://'.`

gno.land was unaffected because it declares `https://…` in the meta tag, so only
Local broke.

**Fix:** replaced `getUrlPathWithoutProtocol` with `normalizeGnoConnectRpc`, which
always returns a fetchable `http(s)` URL — http(s) inputs are kept as-is, and any
non-http/bare input has its scheme (re)applied: loopback hosts → `http://`, all
others → `https://`.

## Security: runtime-gated loopback trust

Statically trusting loopback origins in production would reintroduce exactly the
risk #836 flagged: any local process holding the port could drive the wallet flow.
Instead of a static allowlist, loopback origins are now trusted **only at
runtime**:

- **Remote (https) gno origins** — trusted unconditionally (team-controlled).
- **Loopback origins** — honored only when the wallet's **currently active
  network** is the one that `chains.json` maps to that origin
  (`http://127.0.0.1:8888` → `dev`). The check reads the active `chainId` from the
  wallet (trusted background), not from attacker-controllable meta tags, and fails
  closed.

### Effect

- A malicious `http://127.0.0.1:8888` page cannot drive the wallet while the user
  is on another network (e.g. mainnet) — the flow is silently dropped.
- Loopback is trusted only when the user has **explicitly opted into the Local
  (`dev`) network**, which is the only scenario where a local tx link is expected.

## Notes / behavioral changes

- A loopback tx link requires the wallet to be **unlocked and on the `dev`
  network**. While locked, `getNetwork()` returns `WALLET_LOCKED`, so the gate
  fails closed; unlock and retry.
- Trust is keyed to the exact `chains.json` gnoUrl (`http://127.0.0.1:8888`).
  Serving gnoweb under a different host/port (e.g. `localhost`) requires adding
  that origin to `chains.json`.
- This intentionally relaxes #836's "loopback only in dev builds" stance and
  replaces it with a tighter runtime, active-network gate.

## Changes

- `common/constants/gno-connect-allowlist.constant.ts` — split trust into static
  https allowlist + loopback→chainId map; add `isLoopbackOrigin` /
  `getLoopbackOriginChainId`.
- `inject/message/methods/gno-connect.ts` — add `normalizeGnoConnectRpc` and
  `getLoopbackGnoConnectChainId`; `isAllowedGnoConnectOrigin` now covers https
  origins only.
- `inject/message/command-handler.ts` — runtime active-network gate for loopback
  origins before honoring gnoconnect data.

### DIFF
--- a/packages/adena-extension/src/inject/message/command-handler.ts
+++ b/packages/adena-extension/src/inject/message/command-handler.ts
@@ -21,6 +21,7 @@ import {
   GnoConnectInfo,
   GnoMessageInfo,
   isAllowedGnoConnectOrigin,
+  isLoopbackGnoConnectTrusted,
   parseGnoConnectInfo,
   parseGnoMessageInfo,
 } from './methods/gno-connect';
@@ -145,17 +146,20 @@ export class CommandHandler {
     }
 
     // Runtime trust gate for loopback origins: only honor a local node's
-    // gnoconnect declarations when the wallet is currently on the matching
-    // network. Prevents an arbitrary process holding the local port from driving
-    // the wallet flow when the user is on another network.
+    // gnoconnect declarations when the meta-declared chainId matches the one the
+    // origin is permitted to act as AND that chainId is the wallet's active network.
+    // This prevents an arbitrary process holding the local port from
+    // switching the wallet to a foreign network or driving the flow while the
+    // user is on another network.
     if (isLoopbackOrigin) {
       const networkResponse = await executor.getNetwork();
       const activeChainId =
         networkResponse.type === WalletResponseSuccessType.GET_NETWORK_SUCCESS
           ? networkResponse.data?.chainId
           : undefined;
-      if (activeChainId !== loopbackChainId) {
-        console.info('Loopback gnoconnect origin is not the active network; ignoring');
+
+      if (!isLoopbackGnoConnectTrusted(loopbackChainId, gnoConnectInfo.chainId, activeChainId)) {
+        console.info('Loopback gnoconnect origin is not trusted for the active network; ignoring');
         return;
       }
     }
--- a/packages/adena-extension/src/inject/message/command-handler.ts
+++ b/packages/adena-extension/src/inject/message/command-handler.ts
@@ -16,6 +16,7 @@ import {
 } from './commands/encrypt';
 import { clearPopup } from './commands/popup';
 import {
+  getLoopbackGnoConnectChainId,
   GnoArgumentInfo,
   GnoConnectInfo,
   GnoMessageInfo,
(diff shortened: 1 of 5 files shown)


packages/adena-extension/src/inject/message/command-handler.ts
```diff
@@ -137,6 +144,22 @@ export class CommandHandler {
...
+    // gnoconnect declarations when the wallet is currently on the matching
+    // network. Prevents an arbitrary process holding the local port from driving
+    // the wallet flow when the user is on another network.
+    if (isLoopbackOrigin) {
+      const networkResponse = await executor.getNetwork();
+      const activeChainId =
+        networkResponse.type === WalletResponseSuccessType.GET_NETWORK_SUCCESS
+          ? networkResponse.data?.chainId
+          : undefined;
+      if (activeChainId !== loopbackChainId) {
```
junghoon-vans: The loopback gate only checks that the current wallet network matches the chain mapped from the page origin, but it does not verify that the gnoconnect meta `chainId` is the same chain. If a local page is served from `http://127.0.0.1:8888` while the wallet is already on `dev`, it can still declare another `gnoConnectInfo.chainId` and this code will pass the gate, then call `switchNetwork(gnoConnectInfo.chainId)` below. Please also require `gnoConnectInfo.chainId === loopbackChainId` before honoring loopback gnoconnect data, and add a regression test for `origin=127.0.0.1:8888`, active network `dev`, meta chainId `gnoland1` being rejected.

junghoon-vans (changes requested):
I found one security-relevant gap in the loopback runtime gate. Please check the inline comment.

packages/adena-extension/src/inject/message/command-handler.ts
jinoosss: b468d6ae203eaefaa1ad75f337eba0df47193bc9

junghoon-vans (approved):
LGTM. The loopback runtime gate now validates both the meta-declared chainId and the active wallet network against the origin-mapped chainId, and the regression test covers the foreign chainId case.

PR #869, opened 2026-07-09
fix: enhance account addresses popover with copy functionality

jinoosss:
## Problem

In the header's account-address hover popover, each chain row shows a truncated
address and a small copy icon. Only the **icon** was clickable, so copying an
address required hitting a tiny target — an awkward, easy-to-miss hit area.

## Fix

Make the entire right cell — the truncated address (`g1jg8m…jvsqf5`) **and** the
icon — a single clickable copy target.

- Replaced the right-cell `<div>` with a `<button>` that copies on click, so the
  address text itself is now part of the clickable area.
- The full, non-truncated address is copied (`navigator.clipboard.writeText`),
  even though the UI shows the truncated form.
- Copy feedback (copy → check icon, auto-reverting after 2s) is now tracked per
  row and shared across the whole cell, so clicking either the address or the
  icon gives the same confirmation.
- Extracted each row into a `ChainAddressRow` component to hold its own copy
  state.
- Kept the reusable `CopyIconButton` atom untouched (it is a fixed-size icon used
  across many screens); the copy logic is handled locally in the popover instead.
- Added an `aria-label` on the button for accessibility.

## Changes

- `account-addresses-popover.tsx` — new `ChainAddressRow` with per-row copy state;
  the right cell is now a clickable button wrapping the address + icon.
- `account-addresses-popover.styles.ts` — reset the `.chain-right` button
  (padding/border/background) and add `cursor: pointer` + icon sizing; extend the
  hover/`copied` highlight to the address and icon.

### DIFF
--- a/packages/adena-extension/src/components/pages/router/top-menu/account-addresses-popover/account-addresses-popover.styles.ts
+++ b/packages/adena-extension/src/components/pages/router/top-menu/account-addresses-popover/account-addresses-popover.styles.ts
@@ -35,7 +35,8 @@ export const ChainRow = styled.div`
   gap: 8px;
   transition: color 0.15s ease;
 
-  &:hover {
+  &:hover .chain-right,
+  .chain-right.copied {
     .address {
       color: ${getTheme('neutral', '_1')};
     }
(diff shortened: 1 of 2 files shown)


junghoon-vans (approved):
Reviewed the header account-address copy-area change. The right cell now copies the full address, changed files lint cleanly, and GitHub unit/build checks are passing. LGTM.

PR #872, opened 2026-07-10
feat: support custom HD derivation paths when creating accounts

jinoosss:
## Summary

Adena previously derived accounts only from paths ending in `.../0'/0/N` — the
account', change, and address-index segments were fixed. This PR lets users edit
the derivation path when adding accounts on the **Select Accounts** step (seed-phrase
import and Ledger), while keeping `m/44'/118'` fixed. `account'`, `change`, and
`addressIndex` become editable, and a single account at the entered path can be
added alongside the enumerated list.

## Scope

- Seed-phrase (12/24 word) import → Select Accounts
- Ledger → Select Accounts
- Private key import is out of scope (a raw key has no BIP44 derivation)

### adena-module
- **`hd-path.ts`** — `HdPath { account, change, addressIndex }`, `toHdPath`,
  `formatHdPath`, and `generateKeyPairByHdPath` (custom Slip10 derivation with a
  fixed `44'/118'` prefix). Keyring methods now accept `number | HdPath`; a bare
  number keeps its legacy meaning (address index), so existing accounts derive
  byte-identical keys.
- **`ledger-signer.ts`** — `FullPathLedgerSigner` re-implements tm2's `LedgerSigner`
  with a full path and is injected via `Wallet.fromSigner`, since
  `Wallet.fromLedger` only accepts an address index.
- Accounts carry optional `accountIndex` / `changeIndex` (absent ⇒ 0); the full
  path is threaded through Gno and Cosmos signing.

### adena-extension
- **`HDDerivationPathBox`** — inline editor (account'/change/addressIndex inputs,
  live derived-address preview, `Account already exists` error). Per-field clamping:
  account'/index `0..2^31-1`, change `0..1`.
- **Select Accounts** — the list stays interactive next to the editor; the entered
  path adds one account, deduplicated against the selection and existing accounts,
  ordered by `(account', change, addressIndex)`; a path already in the list shows
  checked + locked.
- `AccountDataModelV021` gains `accountIndex` / `changeIndex` (optional; no data
  migration needed).

## Behavior / hardening
- Address is derived only once every segment is filled.
- Existing accounts are never duplicated on submit (defense in depth).
- HD keyring is reused via content comparison of the mnemonic entropy.
- Ledger accounts are numbered sequentially so names stay unique.

## Backward compatibility
- Bare-index derivation is byte-identical to the previous tm2 path (unit-tested),
  so existing account addresses are unchanged.

## Known limitations / follow-ups
- `tm2-js-client` only varies the address index, so adena had to reimplement path
  derivation and the Ledger signer. Extending tm2-js-client upstream (see doc)
  would let us drop these workarounds.

## Testing
- adena-module: `hd-path.spec.ts` (derivation parity + custom paths), account model
  round-trip; full module suite green.
- [x] Manual: import + Ledger — set path, preview, add; duplicate/lock/order checks.


junghoon-vans (approved):
Reviewed. Overall looks good and checks are passing. One minor recommendation: add central HD path validation in adena-module as defense-in-depth, so direct module usage or malformed persisted data cannot pass invalid account/change/addressIndex values into derivation.

PR #874, opened 2026-07-10
fix: prevent deposits to Session Account addresses

jinoosss:
## Summary

A Session Account address can never receive tokens — funds must be sent to the
Master Account instead. This PR removes every UI affordance that could lead a
user to deposit into a session address and clarifies the intended flow.

### Header
- Removed the copy action for session accounts. The copy icon no longer copies
  the address; hovering it now shows a warning popover instead:
  _"Do not send tokens to your Session Account. Use the Master Account's address
  to receive tokens."_
- Non-session accounts keep the existing per-chain address copy popover.

### Deposit entry points
- Disabled the **Deposit** button for session accounts on both the Wallet Main
  and Token Details screens.

### Account lists & details
- Hid the session account address (and copy icon) in the side menu account list.
- Hid the QR code and address block in the Account Details page for session
  accounts.

### Deposit token page
- Removed the now-deprecated master-address branch (Master label + tooltip).
  Session accounts can no longer reach this page, so the special-case UI was
  dead code.

### Fix
- `MainActionButton` still applied its `:hover` background while `disabled`
  because the `:disabled` rule did not override the background. Scoped the hover
  to `:not(:disabled)` so disabled buttons (e.g. Deposit on a session account)
  no longer react to hover.

### DIFF
--- a/packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
+++ b/packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
@@ -4,8 +4,6 @@ import { WarningPopoverWrapper } from './session-address-warning-popover.styles'
 
 // A SessionAccount address can never receive tokens, so the header copy affordance
 // is replaced by this warning instead of copying an address on hover.
-const SESSION_DEPOSIT_WARNING = `Do not send tokens to your Session Account. Use the <b>Master Account’s</b> address to receive tokens.`;
-
 interface SessionAddressWarningPopoverProps {
   open: boolean;
   positionX: number;
--- a/packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
+++ b/packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
@@ -1,0 +1,41 @@
+import { Portal } from '@components/atoms';
+import React from 'react';
+import { WarningPopoverWrapper } from './session-address-warning-popover.styles';
+
+// A SessionAccount address can never receive tokens, so the header copy affordance
+// is replaced by this warning instead of copying an address on hover.
+const SESSION_DEPOSIT_WARNING = `Do not send tokens to your Session Account. Use the <b>Master Account’s</b> address to receive tokens.`;
+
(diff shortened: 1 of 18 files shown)


packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
```diff
@@ -0,0 +1,41 @@
+import { Portal } from '@components/atoms';
+import React from 'react';
+import { WarningPopoverWrapper } from './session-address-warning-popover.styles';
+
+// A SessionAccount address can never receive tokens, so the header copy affordance
+// is replaced by this warning instead of copying an address on hover.
+const SESSION_DEPOSIT_WARNING = `Do not send tokens to your Session Account. Use the <b>Master Account’s</b> address to receive tokens.`;
```
junghoon-vans: This constant is unused, so the changed-file ESLint check fails with `@typescript-eslint/no-unused-vars` (and the template literal also violates the single-quote rule). Please either remove the constant or render from it instead of duplicating the same text inline below.

junghoon-vans (changes requested):
Blocking issue: the NFT deposit entry point still allows Session Accounts to reach the Deposit page. `packages/adena-extension/src/pages/popup/wallet/nft/index.tsx:82-90` still navigates to `RoutePath.Deposit` through `Deposit NFT` (`packages/adena-extension/src/components/pages/nft/nft-header/nft-header.tsx:18-20`). This leaves a remaining deposit affordance for Session Accounts, and after this PR removes the Master-address label/tooltip from the Deposit page, that route can show the Master address under the current Session Account nickname. Please hide/disable the NFT deposit option for Session Accounts and also guard `moveDepositPage` or the Deposit route itself so the page cannot be reached directly in a Session Account context.

packages/adena-extension/src/components/pages/router/top-menu/session-address-warning-popover/session-address-warning-popover.tsx
jinoosss: 6670750078384ec545a9948146a3ef8bbf269ba9

junghoon-vans (approved):
LGTM. I verified the remaining Session Account deposit entry points are closed, the unused warning constant is removed, and focused lint/tests pass.

PR #875, opened 2026-07-20
refactor: source chain data from chains.json and support the topaz network

jinoosss:
## Summary

The chain registry hardcoded Gno and AtomOne entries in `adena-module`, duplicating the network definitions the extension already ships in its `chains.json` resources. Keeping the two in sync was manual and error-prone.

`createChainRegistry()` now takes chains and network profiles as options, so the app owns the chain data and the module owns only the registry mechanics. With the JSON resources as the single source of truth, this also renames the `test-13` testnet to `topaz` and adds a storage migration for existing installs.

## Changes

**Chain registry**
- `createChainRegistry()` accepts injected `chains`, `gnoNetworkProfiles`, and `atomoneNetworkProfiles`
- Removed the module's `chain-registry/entries/` directory
- Added `chain-utils.ts` on the extension side, holding the Chain definitions and mapping `chains.json` entries into `NetworkProfile`s
- Wired through both registry construction sites: `AdenaProvider` and `InjectCore`

**Network resources**
- Renamed the `test-13` testnet to `topaz`, pointing at the new endpoints
- Unified resource field names across both JSON files: `isMainnet` → `main`, `restUrl` → `apiUrl`
- Typed the optional URL fields on `NetworkMetainfo` as optional, matching resources where not every network declares an indexer, gnoweb, api, or explorer URL

**Storage migration (v022)**
- Repoints the current chain and network id to `topaz`
- Refreshes the default networks from `chains.json`
- Drops leftover `test-13` entries from establish sites, token metainfos, sessions, and GRC721 collections

## Notes

- The registry API change, the field rename, and the call sites do not typecheck independently, so they are committed together to keep every commit bisectable.
- Registry tests in `adena-module` now build on local fixtures rather than production chain data, so they stay stable when a real network is renamed.
- The v020 migration refreshes a hardcoded `test-13` entry that no longer resolves. Its test now asserts the passthrough behavior instead; v022 is what removes the stale network from existing installs.
- The gno-connect allowlist specs were updated for the renamed origin — `https://test13.testnets.gno.land` is no longer trusted, `https://topaz.testnets.gno.land` is.


packages/adena-extension/src/resources/chains/atomone-chains.json
```diff
@@ -11,13 +11,13 @@
     "networkName": "Mainnet",
     "addressPrefix": "atone",
     "rpcUrl": "https://atomone-rpc.allinbits.com",
-    "restUrl": "https://atomone-api.allinbits.com",
+    "apiUrl": "https://atomone-api.allinbits.com",
```
junghoon-vans, line 14: This resource shape was renamed from `isMainnet/restUrl` to `main/apiUrl`, but several runtime consumers still expect the old `AtomoneNetworkMetainfo` shape. For example, `AtomoneNetworkMetainfoMapper` still reads `isMainnet` and `restUrl`, while `states/network.ts` and the wallet-provider fallback cast this JSON directly to `AtomoneNetworkMetainfo[]`. That can leave `isMainnet`/`restUrl` undefined and break AtomOne network selection and LCD endpoints. Please either map `main -> isMainnet` and `apiUrl -> restUrl` at every load boundary, or update the AtomOne metainfo type and all consumers together.

packages/adena-extension/src/common/utils/chain-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export function makeAtomOneNetworkProfiles(resources: NetworkMetainfo[]): CosmosNetworkProfile[] {
+  return resources.map((resource) => {
+    return {
+      ...defaultAtomOneNetworkProfileBase,
+      id: resource.id,
+      chainId: resource.chainId,
+      displayName: resource.networkName,
+      isMainnet: resource.main ?? false,
+      nativeTokenId: makeNativeTokenId(resource.chainId),
```
junghoon-vans: `makeNativeTokenId()` always returns `${chainId}:ugnot`, so AtomOne profiles are registered with native token IDs like `atomone-1:ugnot`. The token registry only has AtomOne native tokens under `uatone`/`uphoton` (for example `atomone-1:uatone`), so native-token lookups for these profiles can fail. Please use an AtomOne-specific native token id such as `${resource.chainId}:uatone` here, and add a fixture/assertion for the generated AtomOne profiles.

packages/adena-extension/src/resources/chains/chains.json
```diff
@@ -15,18 +15,18 @@
     "linkUrl": "https://gnoscan.io"
   },
   {
-    "id": "test-13",
+    "id": "topaz",
```
junghoon-vans: After replacing the default testnet id with `topaz`, the code fallback still searches for `test-13` in `use-network.ts` (`DEFAULT_TESTNET_NETWORK`). Since that id no longer exists, the fallback becomes `CHAIN_DATA[0]`, which is currently mainnet. Please update that fallback to `topaz` or reuse `pickDefaultByMode(..., 'testnet')` so a pre-hydration testnet fallback cannot point at mainnet.

junghoon-vans (changes requested):
I found a few runtime mapping issues around the chain resource shape changes. Please see the inline comments.

packages/adena-extension/src/resources/chains/atomone-chains.json
junghoon-vans, line 14: The new `main/apiUrl` shape is now handled by `AtomoneNetworkMetainfoMapper`, but this JSON is still imported and cast directly to `AtomoneNetworkMetainfo[]` in other runtime paths (`states/network.ts` initializes `ATOMONE_DEFAULT_NETWORKS`, and `wallet-provider.tsx` uses the same cast in the fallback path). Those paths bypass the mapper, so they still produce entries with `isMainnet` and `restUrl` undefined. Please route these direct imports through the mapper as well, or add a shared helper for loading AtomOne resource entries into the metainfo shape.

packages/adena-extension/src/common/utils/chain-utils.ts
junghoon-vans: This is still using the Gno native token helper for AtomOne profiles, so AtomOne network profiles get native token IDs like `atomone-1:ugnot`. The registered AtomOne native token is `uatone` (with `uphoton` as the fee token), so token-registry lookups against this profile native token can fail. Please generate `${resource.chainId}:uatone` for AtomOne profiles and add an assertion for the produced AtomOne profile IDs.

junghoon-vans (changes requested):
Thanks for the updates. The fetch-side AtomOne mapper and the Gno testnet fallback are improved, but two runtime issues still remain.

junghoon-vans (dismissed):
Approved after follow-up review. CI is green; remaining discussion can be handled as non-blocking follow-up.

packages/adena-extension/src/resources/chains/atomone-chains.json
jinoosss: 3f1ad1da6d760dd75fbf68cdf241bf68390fff7b

packages/adena-extension/src/common/utils/chain-utils.ts
jinoosss: 90a9d0a16152e394d4d39b73caee504bd526d1ef

PR #878, opened 2026-07-28
feat(grc20): identify tokens by token path and source them from grc20reg

jinoosss:
## Summary

Treat a GRC20 token by its **token path** `{packagePath}:{symbol}` instead of the bare package path, and make the on-chain **`grc20reg`** registry the single source of truth for GRC20 discovery, metadata, and balances.

A realm can register more than one symbol, and `grc20reg` keys each token by the canonical fqname `{packagePath}.{symbol}` (one token per realm+symbol). Keying the wallet by package path alone collapsed sibling symbols into one entry and made lookups ambiguous. This PR moves the wallet to the token-path model end to end.

## Background

- `grc20reg` registry key = `fqname.Construct(rlmPath, symbol)` = `{packagePath}.{symbol}` (dot).
- Wallet token path (internal convention) = `{packagePath}:{symbol}` (colon).
- Package paths use slashes and never contain `:`; only the trailing `.symbol` segment carries a dot after the last slash, so the two forms convert deterministically.

### Identity & conversion

- New `grc20-token-path` helpers: `toTokenPath`, `parseTokenPath`, `toRegistryKey`, `registryKeyToTokenPath`, `packagePathOfTokenPath`, `isTokenPath`.
- GRC20 `tokenId` is now the token path at every creation/mapping site; equality and dedupe compare `tokenId`.

### Query (list + metadata)

- List GRC20 tokens by paging the `grc20reg` AVL tree via qeval (`GetRegistry().IterateByOffset` + `Size`), **fully replacing** the API-server (`/v1/tokens`) and indexer register-event paths.
- Read metadata from the registry object receiver (`Get(key).GetName/GetSymbol/GetDecimals`), keyed by the fqname.
- Single-token lookup goes through `grc20reg` as well — the qrender/qfile source-parsing fallbacks are removed (GRC721 still uses qfile as `grc20reg` does not cover it).

### Balances

- Resolve GRC20 balances via `grc20reg` (`Get(key).BalanceOf(addr)`, nil → 0).
- Add a batched map query so the wallet-main load fetches many balances in a single qeval instead of one round-trip per token.

### Transfer

- When a chain provides a GRC20 helper realm, call `Transfer(tokenKey, to, amount)` via `MsgCall`.
- Otherwise fall back to a `MsgRun` that resolves the token through `grc20reg` and transfers from the caller.

### Account-held auto-registration

- When the network exposes an API URL, map `/v1/accounts/{address}` assets straight into held GRC20 tokens (identity taken from the response `tokenId`, normalized to the wallet token path) instead of cross-referencing the registry by package path.
- This keeps wallet-main auto-registration, makes it precise for multi-symbol realms, and avoids a full registry scan. Networks without an API URL fall back to indexer-based discovery.

### Config

- New bundled resource `grc20reg.json` maps `chainId → { registryPath, helperPath }`; the previously hardcoded registry path now comes from here.
- `helperPath` is empty until per-chain helper realms are provided, so transfers use the `MsgRun` fallback in the meantime.

### UI

- The add-token browse list loads from `grc20reg` through an infinite query and requests the next page automatically when scrolled near the bottom (guarded against duplicate/exhausted fetches).
- Selecting a list item re-fetches by the exact token path, so multi-symbol realms resolve to the selected token; manual entry by realm path still works.

## Storage migration

- **v023** re-keys existing GRC20 tokens in `ACCOUNT_TOKEN_METAINFOS` from `pkgPath` to `{pkgPath}:{symbol}`.
- No schema change; other token types are untouched.

## Notes / follow-ups

- `helperPath` is empty for all chains today → transfers run the `MsgRun` fallback. Fill `grc20reg.json` when helper realms are available to switch to the `MsgCall` path (no code change needed).
- The on-chain qeval expressions (list/balance IIFEs) and the `MsgRun` transfer body are written against the current `grc20reg` / gnoswap-helper conventions but have not yet been verified end to end against a live node; worth one manual pass on deploy.


packages/adena-extension/src/hooks/use-token-metainfo.tsx
```diff
@@ -94,7 +94,9 @@ export const useTokenMetainfo = (): UseTokenMetainfoReturn => {
         ? grc20Tokens.filter(
             (token) =>
               !fetchedTokenMetainfos.find(
-                (meta) => isGRC20TokenModel(meta) && meta?.pkgPath === token?.pkgPath,
+                // Dedupe by token path (`{packagePath}:{symbol}`), not pkgPath, so
+                // two tokens from the same realm are treated as distinct.
+                (meta) => isGRC20TokenModel(meta) && meta?.tokenId === token?.tokenId,
```
junghoon-vans, line 102: This dedupe now correctly uses `tokenId`, but the downstream metadata maps in this hook are still keyed by `pkgPath` (`tokenMetaMap[current.pkgPath]` and `makeTokenKeyByPackagePath(token.pkgPath)`). For a realm that registers multiple GRC20 symbols, those entries still collapse/overwrite each other when transaction history calls `getTokenAmount(transaction.amount)` or looks up `tokenLogoMap[transaction.amount.denom]`, because the history response only carries `amount.denom` and not the token path. Could we carry the token path through the history mapping (or otherwise key the GRC20 metadata/logo maps by `tokenId`) so the new identity model is applied consistently end-to-end?

junghoon-vans (commented):
Left one comment about carrying the new GRC20 token-path identity through the remaining metadata/history formatting path.

packages/adena-extension/src/hooks/use-token-metainfo.tsx
junghoon-vans: Clarifying this based on the history path: if `amountIn.denom` / `amountOut.denom` already come back as the Adena token path, the issue is specifically at `use-make-transactions-with-time.ts:79` and `:86`.

Those lines call `getTokenAmount(transaction.amount)` and `tokenLogoMap[transaction.amount.denom]`, while `use-token-metainfo.tsx` still builds the GRC20 lookup maps by `pkgPath` (`tokenMetaMap[current.pkgPath]` and `makeTokenKeyByPackagePath(token.pkgPath)`). So a token-path denom like `gno.land/r/demo/foo20:FOO` will not resolve against maps keyed by `gno.land/r/demo/foo20`. We should key the GRC20 metadata/logo maps by `tokenId` (or normalize the history denom before lookup) so history amount/logo formatting uses the same token-path identity.

packages/adena-extension/src/repositories/transaction/transaction-history-indexer.ts
```diff
@@ -112,7 +120,9 @@ export class TransactionHistoryIndexerRepository implements ITransactionHistoryI
       const callTx = tx as TransactionResponse<MsgCallValue>;
       const firstMessage = callTx.messages?.[0];
       const isCallerSelf = firstMessage?.value?.caller === address;
-      return isCallerSelf ? mapVMTransaction(callTx) : mapReceivedTransactionByMsgCall(callTx);
+      return isCallerSelf
+        ? mapVMTransaction(callTx, this.grc20HelperPath)
```
junghoon-vans, line 136: Potential blocker once `helperPath` is enabled: this mapper can decode helper-routed transfers after they are fetched, but the token-detail query above still calls `makeGRC20TransactionHistoryQuery(address, packagePath)`, which filters `pkg_path == token.pkgPath`. Helper transfers have `pkg_path == helperPath` and carry the token key in `args[0]`, so they may be excluded before this mapping runs. Could we include helper `Transfer(tokenKey, to, amount)` calls matching the selected token key/token path?

junghoon-vans (commented):
Left one follow-up blocker note for token-detail history when GRC20 transfers are routed through the helper realm.

packages/adena-extension/src/hooks/use-token-metainfo.tsx
jinoosss: Addressed in 5f01d5ef. The GRC20 lookup maps in `use-token-metainfo` are now keyed by `tokenId` (token path) as the primary key, with a `pkgPath` fallback key for transaction history (which carries only the realm path). So both a token-path denom like `gno.land/r/demo/foo20:FOO` and a bare `pkgPath` resolve, and sibling symbols from one realm no longer collapse. `getTokenAmount` / `tokenLogoMap[amount.denom]` at use-make-transactions-with-time.ts:79,86 resolve correctly either way.

packages/adena-extension/src/repositories/transaction/transaction-history-indexer.ts
jinoosss: Fixed in e2e3540e. The per-token history now threads the token path down the chain: the indexer query adds a branch that matches `helperPath.Transfer(tokenKey, ...)` by the selected token key (registry fqname), in addition to the direct `pkg_path == packagePath` branch. The token-detail page passes the token path (tokenId), the API repo derives the realm path for its endpoint, and the mapper decodes the helper args. With no helper configured (helperPath empty) the query is unchanged.

jinoosss: Follow-up in aae5fad6: per your note, the per-token history query now fetches **both** the token realm and the helper branch, and the mapper identifies the transfer from the GRC20 **`Transfer` event** rather than message args. Direct, helper-routed, and MsgRun transfers all emit the same event with `token` (Token.ID() = `{packagePath}.{symbol}.{sequence}`), `from`, `to`, `value`; the mapper strips the trailing sequence to the token path and reads from/to/value from the event (message-args parsing kept only as a fallback).

junghoon-vans: Thanks, the event-based mapping covers the helper/direct decoding part. One remaining edge for multi-symbol realms: `makeGRC20TransactionHistoryQuery` still fetches the direct branch by `pkg_path == packagePath` + party only. If the same realm has `FOO` and `BAR`, opening `FOO` details will also fetch direct `BAR` transfers for the same address; the mapper will preserve `amount.denom = ...:BAR`, but it is still included in the `FOO` history list. Could we filter the mapped results by the selected `tokenPath` (or add an event-token filter to the direct branch) so per-token history is scoped to the exact token path?

jinoosss: Scoped in 289fba9f. After mapping, `fetchGRC20TransactionHistoryBy` (indexer) filters rows to the selected token: it keeps those whose event-decoded `amount.denom` equals the token path, or the realm path when no event was available (that fallback cannot disambiguate symbols). So opening FOO no longer lists direct BAR transfers from the same realm.

Note: the API history path uses `/grc20-token/{packagePath}/transactions` and its `amount.denom` is backend-provided (format not guaranteed to be the token path), so I left it to the backend to scope per token rather than risk dropping rows with a client-side filter. Happy to add the same filter there once the API denom format is confirmed.

junghoon-vans (dismissed):
Reviewed the latest updates. The token-path history scoping issue is addressed, and I don't see any remaining blockers from my side.

junghoon-vans (dismissed):
Re-reviewed the latest helper path updates. No blockers from my side.

junghoon-vans (dismissed):

PR #882, opened 2026-08-26
perf: cut redundant chain RPC polling across the wallet

jinoosss:
## Problem

An idle wallet popup with five accounts issued roughly **240 RPC requests per minute** before the user touched anything. The bottleneck was not the RPC layer but the polling schedule around it: overlapping timers, a 3-second full transfer-history rescan, and several queries fanning out one request per account, per token, or per loaded page. No RPC method, endpoint, or response handling changes — this is entirely about *when* and *how often* existing calls are made.

## Where the traffic came from

| Interval | Source | Cost per tick |
|---|---|---|
| 3s | Gno balances (`useTokenBalance`) | 2 requests |
| 3s | `accountNativeBalanceMap` | 1 per account |
| 3s | `wallet-main` timer | 3 API calls + duplicate balance refetch |
| 3s | Transaction history | 1 per loaded page |
| 5s | History screen timer | duplicate of the above |
| 5s | Session revocation watcher + session chain data | the same `getSession`, twice |
| 10s | Cosmos balances | 1 per token, per chain |
| per navigation | `checkNetworkState()` | 1 `/health` |

On top of that, the global `QueryClient` had no defaults, so every query ran with `staleTime: 0` and `retry: 3` — each navigation re-issued the screen's entire RPC set, and a struggling node received three times the traffic.

## Changes

**Removed duplicated balance polling.** The `wallet-main` timer no longer refetches balances (the query already had its own `refetchInterval`), and Gno balance polling relaxes from 3s to 5s.

**Stopped rescanning transfer history every 3 seconds.** `updateAllTokenMetainfos()` is token *discovery* — three REST calls on API-backed networks, or an unbounded indexer GraphQL scan plus a full grc20 registry walk (up to 20 pages × 2 qeval) without one. It now runs as a 60s safety net; `initTokenMetainfos` still does a full pass whenever the account or network changes. `fetchTransferTokens` also requested `/v1/accounts/{address}` twice in a single `Promise.all`; the repository now shares the in-flight request per address.

**Capped history polling.** `useInfiniteQuery` refetches every loaded page per tick, and two timers were driving it. The query's `refetchInterval` is gone, the screen's single 5s timer remains, and `refetchTransactions` narrows to page 0 — only the newest page can gain transactions.

**Merged the duplicate session poll.** Both session pollers now share one cache key: the Session Overview query owns the poll, and the revocation watcher reads through it with a stale window just under the interval. The revoke *recheck* still bypasses the cache — confirming a session is really gone must not be answered from a cached read.

**Added global query defaults.** `staleTime: 10s`, `refetchOnWindowFocus: false`, `retry: 1` with capped backoff. `refetchOnMount` is deliberately left at its default so genuinely stale data still refreshes; the stale window is what collapses the burst. The `/health` probe, which bypassed staleTime with an unconditional `refetch()` on every route change, is now `refetchQueries({ stale: true })`.

**Collapsed the per-item fan-outs.**
- *Account balance map → on demand.* The largest steady-state win. It cost one RPC per account and was mounted as a header prefetch, polling on every wallet screen for the life of the popup while rendered nowhere unless the side menu was open. A Recoil counter now tracks how many surfaces render it; the side menu passes its accounts only while open, the accounts screen registers by being mounted, and the header prefetch is removed.
- *Cosmos balances → one request per chain.* `getTokenBalances` now uses the existing `getAllBalances` instead of one `by_denom` call per token.
- *NFT queries → real cache windows.* Per-collection `BalanceOf` had `staleTime: 0` with `refetchOnMount: true` callers, replaying the fan-out on every mount. It and the per-collection token list both move to 30s.


packages/adena-extension/src/hooks/wallet/transaction-history/use-transaction-history-page.ts
```diff
@@ -98,9 +99,18 @@ export const useTransactionHistoryPage = ({
...
-  const refetchTransactions = (options?: RefetchOptions): void => {
-    refetch(options);
-  };
+  // Only the newest page can gain transactions, so a poll refetches page 0 and
+  // leaves the older pages the user scrolled into alone. `fetchNextPage` still
+  // appends normally, and an explicit caller can override `refetchPage`.
+  const refetchTransactions = useCallback(
+    (options?: RefetchOptions): void => {
+      refetch({
+        refetchPage: (_page: unknown, index: number): boolean => index === 0,
```
junghoon-vans: Could we avoid refetching only page 0 here? With cursor pagination, adding a new transaction changes page 0’s boundary. For example, if page 0 previously ended at T20 and page 1 started at T21, a new transaction makes the refreshed page 0 contain N..T19 while the cached page 1 still starts at T21, so T20 disappears from the combined list. `dedupeAndSortTransactions` cannot recover it because the missing boundary item is never fetched again. Please refetch the loaded pages, or otherwise rewind/rebuild subsequent pages when page 0’s cursor changes, so polling cannot drop an existing transaction.

junghoon-vans (commented):
I found one cursor-pagination correctness concern; details are in the inline comment.

packages/adena-extension/src/hooks/use-account-list-infos.ts
```diff
@@ -43,11 +44,18 @@ const buildAccountListInfos = async (
...
+// two account-count-proportional RPC fan-outs in the wallet.
 export const useAccountListInfos = (
   accounts: Account[],
 ): UseQueryResult<SideMenuAccountInfo[], unknown> => {
   const { accountNames } = useAccountName();
   const chain = useChain();
+
+  useAccountBalanceMapDemand(accounts.length > 0);
+
   const { accountNativeBalanceMap } = useTokenBalance();
```
aronpark1007: Not introduced by this PR, but since this change is about reducing polling pressure, could we split the account-native-balance-map query out of `useTokenBalance`?

This hook only needs `accountNativeBalanceMap`, but `useTokenBalance` also mounts the current-account Gno/Cosmos balance pollers. A dedicated hook for the account map would keep this demand gating while avoiding unrelated polling when Accounts or the side menu is opened.

packages/adena-extension/src/hooks/wallet/transaction-history/use-transaction-history-page.ts
junghoon-vans: Confirmed fixed in 3d5f0e0 and retained in the current head. `refetchTransactions` now refetches every loaded page, allowing React Query to rebuild subsequent cursors from the fresh page 0 result. This resolves my concern.

packages/adena-extension/src/hooks/use-account-list-infos.ts
jinoosss: Fixed in 39ebcd09. The map now lives in its own `useAccountNativeBalanceMap` hook.

One note on where this bites: the side menu only renders under HomeMenu, where the wallet is locked and those queries are disabled anyway. 
The real case is the Accounts screen -> wallet-main is unmounted there, so useAccountListInfos was the only observer keeping both balance pollers alive.

aronpark1007 (approved):

junghoon-vans (approved):

PR #883, opened 2026-08-28
Never present an unfetched balance as a real one

jaekwon:
Three related problems, all of which make the wallet lie or go blank when the
browser reports no connection.

### 1. A failed balance fetch was shown as a real balance

`getGnotTokenBalance` ended in `.catch(() => null)`, and its callers render
`` `${balanceAmount || 0}` ``. So a failed RPC call became a confident **0 GNOT** —
identical on screen to an account that genuinely holds nothing. The query
*succeeded*, so there was no error or paused state for the UI to notice.

Cosmos never had this: `fetch-cosmos-balances` propagates errors and sets
`error: true`. During the same outage ATONE correctly showed `-` with a warning
while GNOT showed `0`.

A request that doesn't complete now rejects. `null` is kept for the one thing it
legitimately means: a response that isn't an integer.

Three follow-on fixes so the error reaches the screen: `errorNetworkIds` was
built from Cosmos results only, so a Gno failure had nowhere to go; the headline
balance bypassed it and kept showing a stale figure; and the send flow awaited
`fetchBalanceBy` unguarded, so it now degrades instead of breaking.

### 2. Wallet state was gated on connectivity

The popup and the setup page decide what to render from two `chrome.storage`
reads. Both used react-query's default `networkMode: 'online'`, which *pauses* a
query while the browser reports offline — it never runs, never errors, status
stays `loading`. Both screens render an empty container in exactly that state, so
an offline browser gets a blank popup and a setup page showing only the logo,
with nothing in the console because nothing threw.

Wallet existence and lock state live on disk. They're now `networkMode: 'always'`.

### 3. Nothing ever said "offline"

No screen read `navigator.onLine` or react-query's `onlineManager`. Being offline
was expressed only as data that never arrived. There's now a bar saying so, and
saying what's still trustworthy: accounts and addresses are local and valid, chain
data is paused.

**Retry overrides the signal**, which matters more than it sounds: `navigator.onLine`
is the OS's "is there any route" answer, and a wedged Chrome network notifier
reports offline on a perfectly connected machine — every paused query then hangs
until the browser restarts. The bar stays visible while the browser still claims
offline, so the override is never mistaken for a restored connection, and it's
dropped on the next real outage.

### Deliberately not changed

- The same swallow exists twice for GRC20 (`wallet-balance.ts:72,110`). Left
  alone: `null` is overloaded there (it also means "not an integer") and
  `getGRC20RawBalance` uses a catch for registry→direct-realm *fallback*, which is
  legitimate control flow. Same class of bug, but I couldn't verify a fix.
- `router/web/index.tsx`'s final branch renders a bare `<Route>` outside
  `<Routes>`, which draws nothing. Real, unrelated, separate issue.

### DIFF
--- a/packages/adena-extension/src/App/web.tsx
+++ b/packages/adena-extension/src/App/web.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import { WebRouter } from '@router/web/index';
+import OfflineBanner from '@components/atoms/offline-banner';
 
 import AppProvider from './app-provider';
 import useApp from './use-app';
@@ -8,7 +9,12 @@ import { MemoryRouter } from 'react-router-dom';
 
 const RunApp = (): ReactElement => {
   useApp();
-  return <WebRouter />;
+  return (
+    <>
+      <OfflineBanner />
+      <WebRouter />
+    </>
+  );
 };
 
 const App = (): ReactElement => {
--- a/packages/adena-extension/src/App/web.tsx
+++ b/packages/adena-extension/src/App/web.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement } from 'react';
 import { WebRouter } from '@router/web/index';
-import { OfflineBanner } from '@components/atoms';
 
 import AppProvider from './app-provider';
 import useApp from './use-app';
@@ -9,12 +8,7 @@ import { MemoryRouter } from 'react-router-dom';
 
 const RunApp = (): ReactElement => {
   useApp();
-  return (
-    <>
-      <OfflineBanner />
-      <WebRouter />
-    </>
-  );
+  return <WebRouter />;
 };
 
 const App = (): ReactElement => {
--- a/packages/adena-extension/src/App/web.tsx
+++ b/packages/adena-extension/src/App/web.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement } from 'react';
 import { WebRouter } from '@router/web/index';
-import OfflineBanner from '@components/atoms/offline-banner';
+import { OfflineBanner } from '@components/atoms';
 
 import AppProvider from './app-provider';
 import useApp from './use-app';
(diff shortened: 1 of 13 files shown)


packages/adena-extension/src/components/atoms/offline-banner/index.tsx
jinoosss, line 1: In Adena, components are organized using styled-component.

I will modify it according to the convention 🙏

packages/adena-extension/src/hooks/use-balance-input.ts
jinoosss: Due to the change below, the balance query will return `undefined` if it fails. 
When `currentBalance` is `undefined`, it must be explicitly marked with a `-`.
https://github.com/onbloc/adena-wallet/pull/883/changes#diff-170051e86b2266f7f2d86efc0d0c45243bb49c02679ae5e10486f87f4634e5acR295-R297

packages/adena-extension/src/App/web.tsx
```diff
@@ -8,7 +9,12 @@ import { MemoryRouter } from 'react-router-dom';
 
 const RunApp = (): ReactElement => {
   useApp();
-  return <WebRouter />;
+  return (
+    <>
+      <OfflineBanner />
```
jinoosss: The only time an RPC is called from an actual web route is when creating a session account.
It seems sufficient to perform validation only on that page.

packages/adena-extension/src/pages/popup/wallet/wallet-main/index.tsx
```diff
@@ -257,17 +263,31 @@ export const WalletMain = (): JSX.Element => {
...
-    if (isMainBalanceLoading) return '';
+    if (gnoBalanceUnavailable) return '-';
+    if (mainTokenBalance === null) return '';
     const parsed = BigNumber(mainTokenBalance.value);
     return parsed.isFinite() ? parsed.toFormat() : '-';
   })();
 
   return (
     <Wrapper $dimmed={sessionRevoked}>
+      <OfflineBanner />
```
jinoosss: It appears that the padding in the wrapper is preventing the area from being set as intended.
This creates a padding gap on the actual screen.

jinoosss (commented):
Thank you for bringing this issue to my attention and making the correction 🙏 
I’ve left a few comments, and I’ll add a commit with the changes.

jinoosss (approved):

PR #884, opened 2026-08-29
fix(gnoconnect): say why a loopback TxLink was declined

jaekwon:
The runtime trust gate for loopback origins (#868) fails closed, which is right, but it failed closed in silence: a bare console.info and a return. Four different situations produced the identical no-op — wallet on another network, wallet locked, page declaring a chainId its origin may not act as, and an Adena too old to support loopback TxLinks at all. A user clicking Execute on a local gnodev saw the unlock prompt, then nothing, with no way to tell which of those had happened or what to do about it.

The gate itself is unchanged. getLoopbackGnoConnectRejection now names the reason and isLoopbackGnoConnectTrusted is derived from it, so the message and the decision cannot drift; a test pins that equivalence over every combination. The reason becomes a console.warn and a dismissing notice in the page, naming networks the way the network picker does ("Local (dev)") so the instruction can be followed without translating a chainId.

The notice is diagnostics, not a security surface: any page could draw the same box, so it must never be read as proof that Adena said something. It asks for no input, carries no controls but its own dismiss, and renders page-controlled text with textContent — pinned by an injection test. It tells the page nothing new either, since window.adena.GetNetwork() already reports both the active network and WALLET_LOCKED on request.

### DIFF
--- a/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
+++ b/packages/adena-extension/src/common/constants/gno-connect-allowlist.constant.ts
@@ -53,3 +53,15 @@ const LOOPBACK_ORIGIN_CHAIN_IDS: ReadonlyMap<string, string> = new Map(
 // before honoring the origin's gnoconnect data.
 export const getLoopbackOriginChainId = (origin: string): string | null =>
   LOOPBACK_ORIGIN_CHAIN_IDS.get(origin) ?? null;
+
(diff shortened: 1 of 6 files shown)


jinoosss (commented):
Thank you for improving txlink 🙏 
The code looks good; I'll be back after I test it.

jinoosss (approved):

PR #887, opened 2026-08-31
fix(grc20): follow grc20reg's non-crossing Transfer spec

jinoosss:
## Summary

`grc20reg`'s write wrappers changed from crossing to **non-crossing** functions on master:

```gno
// Call it non-crossing, forwarding your own `cur`:
//	grc20reg.Transfer(0, cur, "gno.land/r/demo/defi/foo20.FOO", to, 100)
func Transfer(_ int, rlm realm, tokenKey string, to address, amount int64) {
	checkErr(MustGet(tokenKey).RealmTeller(0, rlm).Transfer(0, rlm, to, amount))
}
```

`_ int, rlm realm` is the only shape that yields a non-crossing realm parameter, and `CallerTeller()` moved from `*Token` to `*PrivateLedger`. Both of the wallet's GRC20 transfer paths broke as a result. Fixing the MsgRun fallback then surfaced a second, unrelated regression: since [gnolang/gno#6088](https://github.com/gnolang/gno/pull/6088) the auth ante wires `RequireSigForSimulate` to `txCarriesCode`, so MsgRun signatures are verified on the simulate path as well.

### 1. MsgRun fallback follows the registry's documented entry point

`grc20reg.MustGet(key).CallerTeller().Transfer(...)` no longer compiles — that accessor now hangs off the private ledger. The generated run body calls the registry wrapper instead:

```gno
func main(cur realm) {
	grc20reg.Transfer(0, cur, "<tokenKey>", address("<to>"), <amount>)
}
```

Forwarding our own `cur` non-crossing keeps the ephemeral `/e/<caller>/run` realm as the live frame, so `RealmTeller` debits the signing account. A crossing call would mint a fresh `cur` for the registry and spend the registry's own balance. The wrapper also panics through `checkErr`, so a failed transfer reverts rather than reporting success.

### 2. `pearl-1` helper path cleared

`helperPath` pointed at `grc20reg` itself, which worked while `Transfer` was crossing. MsgCall cannot construct a `realm` argument (`convertArgToGno` rejects non-primitive parameter types), so the registry is no longer reachable that way and the chain falls back to MsgRun. Chains with a real helper realm (`gnoland1`, `dev.gnoswap`) are unchanged.

### 3. Gas estimation signs code-bearing transactions

`makeEstimateGasTransaction` sent a pubkey-only placeholder for already-initialized accounts. The node now rejects that for MsgRun with `/std.UnauthorizedError: signature verification failed`. gno's own note:

> The cost: keyless gas estimation no longer works for either message. gnokey signs a second transaction for simulation and is unaffected; other clients that estimate before signing must supply a real signature.

Documents carrying `/vm.m_run` or `/vm.m_addpkg` are now signed for simulation regardless of the `withSignTransaction` flag.

### 4. `@gnolang/gno-js-client` 2.0.3 → 2.0.4

Picks up the current MsgRun proto. `send` and `max_deposit` lost `omitempty` on the chain side, so they must stay present in the sign payload even when empty.

### 5. Token history matches the Transfer event

The per-token history query matched on message shape (`pkg_path` + `func` + `args`), which cannot see a MsgRun — its body is not indexed. It now matches transactions carrying a `Transfer` event for the selected token id:

```graphql
GnoEvent: {
  type: { eq: "Transfer" }
  _and: [
    { attrs: { key: { eq: "token" }, value: { like: "<tokenKey>." } } }
    { attrs: { key: { eq: "from" }, value: { eq: "<address>" } } }
  ]
}
```

Direct, helper-routed and MsgRun transfers all emit the same event, so one branch covers every invocation shape. The event's `token` attr is `Token.ID()` = `{packagePath}.{symbol}.{sequence}` while the wallet's identity is `{packagePath}.{symbol}`, hence the prefix match. MsgRun transactions carrying that event are now mapped as transfers instead of a generic "Message Run" contract interaction.

### DIFF
--- a/packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
+++ b/packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
@@ -12,6 +12,7 @@ import { useCurrentAccount } from '@hooks/use-current-account';
 import { TransactionService } from '@services/index';
 
 import { useIsInitializedAccount } from '../use-get-account-info';
+import { canSignInProcess, resolveStaticCodeMessageGasUsed } from './code-message-static-gas';
 import { useGetGasPrice } from './use-get-gas-price';
 
 export const GET_ESTIMATE_GAS_KEY = 'transactionGas/useGetEstimateGas';
@@ -83,16 +84,11 @@ function modifyDocument(document: Document, gasWanted: number, gasFee: number): 
   };
 }
 
-// Messages the chain authorizes from their signer, and therefore verifies the
-// signature of even on the simulate path (gnolang/gno#6088 wires the ante's
-// RequireSigForSimulate to txCarriesCode). A placeholder tx carrying one of
-// these is rejected with `/std.UnauthorizedError: signature verification
-// failed`, so gas estimation has to sign for real. Keep in sync with
-// txCarriesCode in gno.land/pkg/gnoland/app.go.
-const SIGN_REQUIRED_MESSAGE_TYPES = ['/vm.m_run', '/vm.m_addpkg'];
-
+// The chain verifies signatures on the simulate path for code-bearing messages
+// (gnolang/gno#6088 wires the ante's RequireSigForSimulate to txCarriesCode), so
+// a placeholder tx carrying one is rejected with `/std.UnauthorizedError`.
 function requiresRealSignature(document: Document): boolean {
-  return (document.msgs ?? []).some((msg) => SIGN_REQUIRED_MESSAGE_TYPES.includes(msg.type));
+  return resolveStaticCodeMessageGasUsed(document) !== null;
 }
 
 /**
@@ -192,6 +188,19 @@ export const useGetEstimateGas = (
         return null;
       }
 
+      // Ledger/AirGap cannot sign while the fee is calculated, and the chain
+      // refuses an unsigned simulate for code-bearing messages, so use the
+      // static table instead of a request that can only fail.
+      const staticGasUsed = resolveStaticCodeMessageGasUsed(document);
+      if (staticGasUsed !== null && !canSignInProcess(currentAccount)) {
+        return {
+          gasUsed: staticGasUsed,
+          storageDeposits: EMPTY_STORAGE_DEPOSITS,
+          hasError: false,
+          simulateErrorMessage: null,
+        };
+      }
+
       const tx = await makeEstimateGasTransaction(
         wallet,
         currentAccount,
@@ -207,26 +216,22 @@ export const useGetEstimateGas = (
 
       const nextResult = await transactionGasService
         .simulateTx(tx)
-        .then(
-          (simulateResult): EstimateGasResult => {
-            return {
-              gasUsed: Number(simulateResult.gas_used),
-              storageDeposits: parseStorageDeposits(simulateResult.response_base?.events ?? []),
-              hasError: false,
-              simulateErrorMessage: null,
-            };
-          },
-        )
-        .catch(
-          (e: Error): EstimateGasResult => {
-            // eslint-disable-next-line no-console
-            console.error('[estimate-gas] simulate failed:', e);
-            return {
-              ...ERROR_ESTIMATE_GAS_RESULT,
-              simulateErrorMessage: e?.message || '',
-            };
-          },
-        );
+        .then((simulateResult): EstimateGasResult => {
+          return {
+            gasUsed: Number(simulateResult.gas_used),
+            storageDeposits: parseStorageDeposits(simulateResult.response_base?.events ?? []),
+            hasError: false,
+            simulateErrorMessage: null,
+          };
+        })
+        .catch((e: Error): EstimateGasResult => {
+          // eslint-disable-next-line no-console
+          console.error('[estimate-gas] simulate failed:', e);
+          return {
+            ...ERROR_ESTIMATE_GAS_RESULT,
+            simulateErrorMessage: e?.message || '',
+          };
+        });
 
       // Keep the higher gasUsed across refetches for the same transaction.
       // Treats an errored/zero response as the lowest value, so a transient
--- a/packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
+++ b/packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
@@ -83,10 +83,26 @@ function modifyDocument(document: Document, gasWanted: number, gasFee: number): 
   };
 }
 
+// Messages the chain authorizes from their signer, and therefore verifies the
+// signature of even on the simulate path (gnolang/gno#6088 wires the ante's
+// RequireSigForSimulate to txCarriesCode). A placeholder tx carrying one of
+// these is rejected with `/std.UnauthorizedError: signature verification
+// failed`, so gas estimation has to sign for real. Keep in sync with
+// txCarriesCode in gno.land/pkg/gnoland/app.go.
+const SIGN_REQUIRED_MESSAGE_TYPES = ['/vm.m_run', '/vm.m_addpkg'];
+
+function requiresRealSignature(document: Document): boolean {
+  return (document.msgs ?? []).some((msg) => SIGN_REQUIRED_MESSAGE_TYPES.includes(msg.type));
+}
+
 /**
  * Builds a placeholder/signed Tx used only for simulation. The fee amount does
  * not affect the simulated `gas_used` (gasWanted is always DEFAULT_GAS_WANTED),
  * so callers can pass any `gasUsed`/`gasPrice` that yields a valid document.
+ *
+ * `withSignTransaction` forces signing for accounts the node cannot validate
+ * from a pubkey alone; a document carrying a code-bearing message forces it
+ * too, regardless of that flag.
  */
 export const makeEstimateGasTransaction = async (
   wallet: Wallet | null,
@@ -114,7 +130,7 @@ export const makeEstimateGasTransaction = async (
   // master caller. Without this, simulate would reject the placeholder for
   // pubkey-address mismatch (master caller + session pubkey).
   const sessionAddr = isSessionAccount(account) ? await account.getAddress('g') : undefined;
-  if (!withSignTransaction) {
+  if (!withSignTransaction && !requiresRealSignature(modifiedDocument)) {
     return documentToDefaultTx(modifiedDocument, account.publicKey, sessionAddr);
   }
 
--- a/packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
+++ b/packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
@@ -113,7 +113,7 @@ export function mapTransactionEdgeByAddress(
   helperPath?: string,
 ): TransactionInfo {
   if (!transaction?.messages?.length || transaction?.messages?.length > 1) {
-    return mapVMTransaction(transaction, helperPath);
+    return mapVMTransaction(transaction, helperPath, undefined, address);
   }
 
   const message = transaction.messages[0];
@@ -133,9 +133,9 @@ export function mapTransactionEdgeByAddress(
       ) {
         return mapReceivedTransactionByMsgCall(transaction, helperPath);
       }
-      return mapVMTransaction(transaction, helperPath);
+      return mapVMTransaction(transaction, helperPath, undefined, address);
     default:
-      return mapVMTransaction(transaction, helperPath);
+      return mapVMTransaction(transaction, helperPath, undefined, address);
   }
 }
 
@@ -274,6 +274,7 @@ export function mapVMTransaction(
   tx: TransactionResponse<AddPackageValue | MsgRunValue | MsgCallValue>,
   helperPath?: string,
   tokenKey?: string,
+  viewerAddress?: string,
 ): TransactionInfo {
   const firstMessage = getDefaultMessage(tx.messages);
 
@@ -428,6 +429,10 @@ export function mapVMTransaction(
   const runTransfer = getGRC20TransferFromEvent(tx, tokenKey);
   if (runTransfer?.tokenPath) {
     const fromAddress = runTransfer.from || (firstMessage.value as MsgRunValue).caller || '';
+    // The event names both parties, so the direction is only known once the
+    // viewer is: the same transaction is a send for one side and a receive for
+    // the other.
+    const isReceive = !!viewerAddress && runTransfer.to === viewerAddress;
 
     return {
       hash: tx.hash,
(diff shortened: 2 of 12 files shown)


packages/adena-extension/src/repositories/transaction/transaction-history-indexer.queries.ts
```diff
@@ -160,72 +160,57 @@ query getNativeTransactionHistory {
 `;
 
 /**
- * GRC20 transfers (sent or received) for an address, scoped to a single
- * token package. `caller eq address` catches sends; `args eq address`
- * catches the recipient slot of `Transfer(to, amount)`.
+ * GRC20 transfers (sent or received) for an address, scoped to a single token.
+ *
+ * Matched on the emitted `Transfer` event rather than on the message shape.
```
junghoon-vans, line 195: The new token-specific query is event-based, but `makeAllTransactionHistoryQuery` above still filters only message fields and `MsgRun.caller`; it never matches `response.events[].GnoEvent.to`. A recipient of a registry-routed `MsgRun` transfer will therefore be absent from the wallet's general transaction history, even though this query finds it. Please add the event-based recipient predicate to the all-transactions path as well, or otherwise merge this result into that path.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -404,6 +404,40 @@ export function mapVMTransaction(
     };
   }
 
+  // MsgRun: a GRC20 transfer routed through the registry's non-crossing
+  // `Transfer(0, cur, tokenKey, to, amount)` wrapper (the path taken when the
+  // chain has no GRC20 helper realm) carries no func/args, so the emitted
+  // Transfer event is the only description of what moved.
+  const runTransfer = getGRC20TransferFromEvent(tx);
```
junghoon-vans: Selecting the first matching event is unsafe when one transaction emits multiple GRC20 `Transfer` events. This helper has no `tokenKey` parameter, so it can select another token's event first; the later denom filter then drops the requested token's transaction altogether. Please pass the requested token key into the mapper and select the event whose token starts with `${tokenKey}.` (and has a `value` attribute) before mapping.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -404,6 +404,40 @@ export function mapVMTransaction(
...
+  if (runTransfer?.tokenPath) {
```
junghoon-vans, line 453: Please require the `value` attribute before classifying an event as GRC20. GRC721 `Transfer` events also carry a `token` attribute but use `tokenId` instead of `value`; without this guard, an NFT event can be decoded as a zero-amount GRC20 transfer when its derived path happens to resolve. A fixture containing both event shapes would catch this.

junghoon-vans (commented):
I found four correctness concerns; details are inline.

packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
```diff
@@ -114,7 +130,7 @@ export const makeEstimateGasTransaction = async (
   // master caller. Without this, simulate would reject the placeholder for
   // pubkey-address mismatch (master caller + session pubkey).
   const sessionAddr = isSessionAccount(account) ? await account.getAddress('g') : undefined;
-  if (!withSignTransaction) {
+  if (!withSignTransaction && !requiresRealSignature(modifiedDocument)) {
```
junghoon-vans, line 129: Now that `MsgRun` is forced through `createTransaction`, this catch still returns an unsigned placeholder for any signing failure. On nodes enforcing `RequireSigForSimulate` for `MsgRun`, that fallback is rejected again, so Ledger/AirGap or otherwise unavailable-signer accounts cannot obtain a gas estimate for GRC20 transfers. Please surface the signing error or use a supported signer-specific simulation path instead of retrying the known-invalid placeholder.

packages/adena-extension/src/repositories/transaction/transaction-history-indexer.queries.ts
jinoosss: Fixed in 3e6f8a08.

Added `makeGRC20ReceivedTransactionHistoryQuery` (Transfer event with `to == address`) and merged it into `fetchAllTransactionHistoryBy`, deduped by hash and re-sorted newest-first.

It is a second query rather than a branch in the existing one because the event filter lives under `where.response` while the message filter lives under `where.messages`, so they cannot share one `_or`, and I could not verify that a top-level `_or` is supported.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
jinoosss: Fixed in 3e6f8a08.

`getGRC20TransferFromEvent` now takes the requested token key and selects the event whose `token` starts with `${tokenKey}.`. The key is threaded from `fetchGRC20TransactionHistoryBy` through `mapVMTransaction` and `mapReceivedTransactionByMsgCall`, so the transaction is mapped from its own event instead of being dropped by the denom filter.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
jinoosss: Fixed in 3e6f8a08.

The event now has to carry a `value` attribute to be treated as GRC20. Confirmed against master that `p/demo/tokens/grc721` emits `Transfer` with `token`/`from`/`to`/`tokenId` and no `value`, so this is what separates the two.

packages/adena-extension/src/hooks/wallet/transaction-gas/use-get-estimate-gas.ts
jinoosss: Fixed in 684bc680.

Rather than retry a request that can only fail, `useGetEstimateGas` now short-circuits before building the tx when the account cannot sign in-process (Ledger/AirGap) and the document carries a code-bearing message, and takes `gasUsed` from a static table in `code-message-static-gas.ts` keyed by message type (add package / plain MsgRun / the GRC20 transfer MsgRun the wallet generates), so each can be tuned independently. Signing on the estimation path would not work for these accounts anyway: the connector is only acquired later in the Ledger flow, and the query refetches every 5s.

The placeholder fallback stays for non code-bearing messages, where simulate still skips signature verification.

The table values are conservative placeholders — I had no measured `gas_used` to base them on, so they should be replaced with numbers observed from an in-process account.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -404,6 +421,40 @@ export function mapVMTransaction(
...
+    const fromAddress = runTransfer.from || (firstMessage.value as MsgRunValue).caller || '';
+
+    return {
+      hash: tx.hash,
+      height: tx.block_height,
+      logo: runTransfer.tokenPath,
+      type: 'TRANSFER',
+      status: tx.success ? 'SUCCESS' : 'FAIL',
+      typeName: 'Send',
```
aronpark1007: This branch now also handles rows pulled in by `makeGRC20ReceivedTransactionHistoryQuery`, where the current wallet is the `Transfer.to` address. Since `mapVMTransaction` does not know the viewer address, those received MsgRun transfers are still rendered as `Send` in the general history.

Please thread the viewer address into this mapper or use a dedicated received mapper for this path. For example:

```ts
const isReceive = viewerAddress && runTransfer.to === viewerAddress;

return {
  ...base,
  typeName: isReceive ? "Receive" : "Send",
  title: isReceive ? "Receive" : "Send",
  description: isReceive
    ? `From: ${formatAddress(fromAddress)}`
    : `To: ${formatAddress(runTransfer.to)}`,
  valueType: mapValueType(tx.success, isReceive),
};
```

Without that, Pearl MsgRun GRC20 transfers received by this address are included by the new query but displayed with the wrong direction.

jinoosss: Good catch — fixed in 5e55c857.

`mapTransactionEdgeByAddress` now passes the viewer address down to `mapVMTransaction`, and the MsgRun branch flips `typeName`/`title`/`description`/`valueType` when the event's `to` is the viewer, as suggested. `from`/`to`/`originFrom`/`originTo` stay factual either way.

Added `transaction-history-query.mapper.spec.ts` covering both directions of the same transaction (and a GRC721 event staying a contract interaction).

The token-detail path was already correct — `fetchGRC20TransactionHistoryBy` routes non-self callers to `mapReceivedTransactionByMsgCall` — so only the general history needed the viewer.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -404,6 +422,46 @@ export function mapVMTransaction(
     };
   }
 
+  // MsgRun: a GRC20 transfer routed through the registry's non-crossing
+  // `Transfer(0, cur, tokenKey, to, amount)` wrapper (the path taken when the
+  // chain has no GRC20 helper realm) carries no func/args, so the emitted
+  // Transfer event is the only description of what moved.
+  const runTransfer = getGRC20TransferFromEvent(tx, tokenKey);
```
junghoon-vans: The received-history query matches a transaction when any Transfer event has `to == viewerAddress`, but this call still selects the first GRC20-like event because `tokenKey` is undefined. If a MsgRun emits multiple transfers and the viewer is the recipient only of a later event, the mapper can show the wrong token, counterparty, and direction. Please select the event involving the viewer for the general-history path (while keeping tokenKey matching for token-detail history), and add a multi-event regression test.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -404,6 +422,46 @@ export function mapVMTransaction(
...
+  if (runTransfer?.tokenPath) {
+    const fromAddress = runTransfer.from || (firstMessage.value as MsgRunValue).caller || '';
+    // The event names both parties, so the direction is only known once the
+    // viewer is: the same transaction is a send for one side and a receive for
+    // the other.
+    const isReceive = !!viewerAddress && runTransfer.to === viewerAddress;
```
junghoon-vans: This also treats a self-transfer (`from === to === viewerAddress`) as `Receive`. That differs from the existing bank.MsgSend and MsgCall mapping, where the account initiating the transfer is rendered as `Send`. Please make the sender check take precedence (or explicitly define the desired self-transfer semantics) and cover it with a test.

packages/adena-extension/src/repositories/transaction/transaction-history-indexer.queries.ts
```diff
@@ -135,6 +135,36 @@ query getAllTransactionHistory {
...
+export const makeGRC20ReceivedTransactionHistoryQuery = (address: string): string => `
+query getGRC20ReceivedTransactionHistory {
+  getTransactions(
+    where: {
+      success: { eq: true }
+      response: {
+        events: {
+          GnoEvent: {
+            type: { eq: "Transfer" }
+            attrs: {
```
junghoon-vans, line 153: This query is named and consumed as GRC20-only, but it filters only `Transfer` and `to`. GRC721 emits the same `token`/`from`/`to` attributes with `tokenId` instead of `value`, so an NFT receive can also be returned here and then rendered as a generic MsgRun contract interaction. Please require the GRC20 event shape (at least `token` and `value`) or otherwise constrain this query to avoid mixing NFT/custom Transfer events into general history.

junghoon-vans (commented):
Re-reviewed the latest head. The previously reported issues are addressed, but the following edge cases still need attention.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
jinoosss: Fixed in 195dcc67.

Event selection is now `tokenKey` first, then the transfer the viewer took part in (`from` or `to` === viewer), then the first GRC20 event. Token-detail history keeps matching by token key; the general-history path resolves by viewer.

Added a regression test with two transfers in one MsgRun where the viewer is only party to the second.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
jinoosss: Fixed in 195dcc67.

`isReceive` now also requires `from !== viewerAddress`, so a self-transfer renders as Send like the bank.MsgSend and MsgCall paths. Covered by a test.

packages/adena-extension/src/repositories/transaction/transaction-history-indexer.queries.ts
jinoosss: Fixed in 195dcc67, but on the client rather than in the query.

`fetchAllTransactionHistoryBy` now drops rows without a GRC20 `Transfer` to the viewer (`hasGRC20TransferTo`, sharing the same `token` + `value` shape check as the mapper) before merging, so NFT receives no longer reach the general history.

I did not add the constraint to the filter itself because expressing "has an attr with key `value`" needs an `attrs` entry with no `value` predicate, and I could not verify that the indexer accepts it — if it does not, the query fails outright and takes the whole history with it. Constraining by `GnoEvent.pkg_path` is not an option either: `chain.Emit` attributes the event to the token realm, not to `p/demo/tokens/grc20`. Happy to move the check server-side if you can confirm the key-only attr filter is supported.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
```diff
@@ -190,7 +228,7 @@ export function mapReceivedTransactionByMsgCall(
 
   // Prefer the GRC20 Transfer event (token/from/to/value); fall back to parsing
   // message args (with the helper arg offset) when it is absent.
-  const eventInfo = getGRC20TransferFromEvent(tx);
+  const eventInfo = getGRC20TransferFromEvent(tx, { tokenKey });
```
junghoon-vans: The viewer-aware event selection is only threaded through `mapVMTransaction`. `mapTransactionEdgeByAddress` still routes a non-self `exec` transfer to `mapReceivedTransactionByMsgCall`, and this call omits `viewerAddress`, so it still selects the first GRC20 event. When the received-history query returns an `exec` transaction whose viewer-target Transfer is not first, the receive row can show the wrong token/counterparty. Please thread `viewerAddress` through this receive mapper as well and add the analogous multi-event `exec` regression test.

junghoon-vans (commented):
Re-reviewed the latest head. The three previous follow-up findings are addressed, but the same event-selection invariant is still missing from the `exec` receive path.

packages/adena-extension/src/repositories/transaction/mapper/transaction-history-query.mapper.ts
jinoosss: Fixed in 3e91e597.

`mapReceivedTransactionByMsgCall` now takes `viewerAddress` and passes it into the event selection, and `mapTransactionEdgeByAddress` supplies it on the `exec` receive route. Token-detail history is unchanged — it still selects by `tokenKey`.

Added the analogous regression test: an `exec` Transfer emitting two GRC20 events where the viewer is party only to the second.

aronpark1007 (approved):

junghoon-vans (approved):
Re-reviewed the latest head (3e91e597). All previously raised history and gas-estimation findings are resolved, focused tests pass, and the required CI checks are green.

PR #895, opened 2026-09-11
feat: add gnoland-1 mainnet and retire pearl-1 and gnoland1

jinoosss:
## Summary

Replace the Beta Mainnet (`gnoland1`) and Pearl (`pearl-1`) networks with the `gnoland-1` Mainnet.

### Network configuration
- **chains.json**: `gnoland-1` (Mainnet), `staging`, `dev`, `gnoswap`. `gnoland1` and `pearl-1` are removed.
- **grc20reg.json**: entries keyed by the new chain ids; the registry path is `gno.land/r/demo/defi/grc20reg/v1`.
- **Defaults**: fresh installs now start on mainnet (`gnoland-1`); the testnet default is `staging`.
- **Account sessions**: `SESSION_UNSUPPORTED_CHAIN_IDS` is emptied, so sessions are enabled on every chain including the new mainnet.
- Renamed every `gnoland1` reference (Recoil defaults, chain icon map, stories, specs) to `gnoland-1`, and dropped the retired gnoconnect origins.

### Storage migration v026
- `gnoland1` → `gnoland-1` and `pearl-1` → `staging` for `CURRENT_CHAIN_ID` / `CURRENT_NETWORK_ID`, so users stay on the same mainnet/testnet side.
- Neither retired chain is a rename, so their chain-scoped state (established sites, token metainfos, sessions, GRC721 collections and pinned packages) is dropped. `NETWORKS` is refreshed from `chains.json`; custom networks are preserved.

## Testing
- `tsc --noEmit` passes.
- New `storage-migration-v026.spec.ts`; older migration specs that asserted `pearl-1` in `chains.json` (and the default network count) now assert the new configuration.

## Notes for reviewers
- `gno.land/r/demo/defi/grc20reg/v1` is not deployed on the current `rpc.gno.land` (still `gnoland1`) or staging, and differs from gnolang/gno#6162 (`r/nt/grc20reg/v0`).
- Multi-version GRC20 registry support follows in a separate PR stacked on this one.

PR #896, opened 2026-09-11
feat(grc20): support several grc20reg registry and grc20 package versions per chain

jinoosss:
## Summary

Stacked on #895. Make GRC20 handling work across several deployed versions of the `grc20reg` registry and the `grc20` package on the same chain, and remove every hardcoded `gno.land/r/demo/defi/grc20reg` / `gno.land/p/demo/tokens/grc20` path.

### Configuration
`grc20reg.json` now describes each chain with:
- `registries`: `grc20reg` realms, newest first (`grc20reg/v1`, then `r/nt/grc20reg/v0`).
- `tokenPackages`: `grc20` packages that emit transfer events, each with its own `transferEvent` schema (event type and `token`/`from`/`to`/`value` attribute keys), since versions may emit differently shaped events.
- `helperPath`: unchanged.

Chains without an entry fall back to `r/nt/grc20reg/v0` / `p/nt/grc20/v0` and the default event shape.

### Behaviour across versions
- **Token listing** pages the registries as one virtual list (newest first) and de-duplicates tokens registered in more than one version.
- **Metadata and balances** resolve each key from the first registry holding it; balance qevals return `(found, int64)` so "not in this registry" is distinguishable from a zero balance. Registries not deployed on a chain are skipped.
- **MsgRun transfers** import every configured registry and transfer through the first one holding the token.
- **Indexer queries** (transfer discovery, received history, per-token history) generate one branch per `grc20` version using that version's `pkg_path`, event type and attribute keys.
- **History mapper** decodes each transfer event with the schema of the package that emitted it.

## Testing
- `tsc --noEmit` passes.
- New specs: `grc20reg-config.spec.ts`, `token.queries.spec.ts`, `transaction-history-indexer.queries.spec.ts`, and multi-registry balance cases in `wallet-balance.spec.ts`.

## Notes for reviewers
- `grc20reg.json` lists only `gno.land/p/nt/grc20/v0` as a token package and no helper realm for `dev.gnoswap`. Transfers emitted by other `grc20` packages are not matched by the indexer queries.


packages/adena-extension/src/services/wallet/wallet-balance.ts
```diff
@@ -140,43 +144,72 @@ export class WalletBalanceService {
...
     }
 
-    for (let start = 0; start < batchable.length; start += GRC20_BALANCE_BATCH_SIZE) {
-      const chunk = batchable.slice(start, start + GRC20_BALANCE_BATCH_SIZE);
+    let pending = batchable;
+    for (const registryPath of this.registryPaths) {
+      if (pending.length === 0) {
+        break;
+      }
+      const resolved = await this.batchGRC20BalancesFrom(
```
notJoon, line 152: question: Would the caller be able to distinguish `{}` for a batch evaluation failure from having no token?

notJoon (dismissed):
left a minor question. otherwise lgtm 👍

packages/adena-extension/src/services/wallet/wallet-balance.ts
jinoosss: Right now the caller can't tell the two apart from the map itself, but the design doesn't rely on that distinction.

An empty result from `batchGRC20BalancesFrom` is treated as "this registry didn't resolve these tokens", which is the same signal as `Get(key) == nil`. `getGRC20TokenBalanceMap` keeps every unresolved path as `pending`, tries the next registry version, and then falls back to per-token `getGRC20RawBalance` (each registry individually, then the realm's `BalanceOf`). Anything still unresolved is simply left out of the map, and `useTokenBalance.fetchBalances` routes those tokens through the existing `fetchBalanceBy` path.

So a batch failure never turns into a wrong balance; it only costs a few extra round-trips for that chunk. A full RPC outage is still surfaced to the UI, since the native balance query in the same `fetchBalances` call doesn't swallow errors and flips the network into the error state.

notJoon: got it. then it should not be a problem

notJoon (approved):
LGTM 👍

PR #899, opened 2026-09-14
feat: adopt tm2-js-client 3.0.0 and gno-js-client 2.2.0 (session_addr, Ledger-compatible sign payload, package approval messages)

jinoosss:
### What type of PR is this?
feature

### What this PR does:

Adopts the latest gno client libraries and removes the local workarounds they replace.

- `@gnolang/tm2-js-client` `2.0.4` → `3.0.0`
- `@gnolang/gno-js-client` `2.0.4` → `2.2.0`

Commits are split by library change so each can be reviewed on its own.

#### 1. `chore: bump tm2-js-client to 2.1.0 and drop local session_addr proto workaround`
- tm2-js-client 2.1.0 ([gnolang/tm2-js-client#284](https://github.com/gnolang/tm2-js-client/pull/284)) syncs `TxSignature` with upstream `std.Signature` and adds `session_addr`.
- Removes `packages/adena-module/src/proto/session/` (`LocalTxSignature`, `encodeGnoTx`, `extractSessionAddress*`), which was a stopgap encoder for that missing field.
- Every call site (`signGnoDocument`, `documentToDefaultTx`, `SessionKeyring` broadcast, `GnoProvider.simulateTx`, transaction hash/encode) now uses `Tx.encode` with `TxSignature.session_addr` directly. An empty `session_addr` is not written to the wire, so master-key transactions encode byte-for-byte as before.

#### 2. `feat: bump gno-js-client to 2.2.0 and support MsgEnablePackage / MsgRejectPackage`
- gno-js-client 2.2.0 ([gnolang/gno-js-client#260](https://github.com/gnolang/gno-js-client/pull/260)) adds `MsgEnablePackage` (`/vm.m_enable_pkg`) and `MsgRejectPackage` (`/vm.m_reject_pkg`) from gnolang/gno#6171.
- Both messages are wired into `decodeTxMessages` / `encodeMessageValue` in `adena-module`, and through the extension transaction pipeline:
  - `EMessageType` / `FUNCTION_NAME_MAP` / `TMessage` and the `RawTxMessageType` union
  - request validators (`validateTransactionMessageOfEnablePackage` / `RejectPackage`), the injection type allow-lists and the `approver` / `sender` address checks
  - `mappedTransactionMessages` / `mappedDocumentMessagesWithCaller` / `mappedRawTxMessages`
  - approve popup renderers, broadcast and multisig broadcast screens, sign-multisig caller extraction and the multisig raw-tx mapper
- Session accounts intentionally remain unable to sign these messages (`session-signing-guard` is unchanged).

#### 3. `feat!: bump tm2-js-client to 3.0.0 and render the sign payload fee in the Ledger-compatible shape`
- tm2-js-client 3.0.0 ([gnolang/tm2-js-client#283](https://github.com/gnolang/tm2-js-client/pull/283)) renders the signature payload fee as `{"amount":[{"amount","denom"}],"gas"}` to match `std.GetSignaturePayload` after [gnolang/gno#6173](https://github.com/gnolang/gno/pull/6173). The Ledger Cosmos app only allows `amount`/`gas`/`granter`/`payer` inside `fee`, so the previous `gas_fee`/`gas_wanted` payload was rejected by the device.
- `signGnoDocument` (HD / private key / Web3Auth / session keyrings) builds the same payload with `parseSignCoin` and the now-exported `sortedJsonStringify`, keeping it bit-equal with the tm2 `Wallet.signTransaction` path that the Ledger keyring uses. The existing parity tests cover this.
- The multisig service parses `gas_fee` with `parseSignCoin` (same rules as `std.ParseCoin`) instead of a local regex.
- The golden signature fixture in `wallet-sign.spec.ts` is regenerated for the new payload.
- gno-js-client 2.2.0 still declares `@gnolang/tm2-js-client: ^2.1.0`, so a root yarn `resolutions` entry pins tm2-js-client to `3.0.0` to keep a single copy installed. It can be dropped once gno-js-client raises its floor.

The approve popup is unaffected: it renders the `Document` (whose fee was already in `{amount, gas}` form); only the bytes handed to the signer change.

### Breaking change / rollout note

Commit 3 changes the signed bytes for every signer. Nodes without gnolang/gno#6173 reject those transactions with `signature verification failed`, while upgraded nodes accept both the legacy and the new payload. This should only be released after the target networks (including `gnoland-1`) run a node with gnolang/gno#6173.

### Verification
- `adena-module`: build OK, 24 suites / 209 tests passed
- `adena-extension`: `tsc --noEmit` OK, 189 suites / 749 tests passed


packages/adena-module/src/utils/messages.ts
```diff
@@ -321,6 +338,30 @@ function encodeMessageValue(message: { type: string; value: any }) {
...
+    case MsgEndpoint.MSG_ENABLE_PKG: {
+      const value = message.value;
+      const msg = MsgEnablePackage.create({
+        approver: value.approver,
+        pkg_path: value.pkg_path,
+        pkg_hash: value.pkg_hash || '',
+        pkg_height: toProtoBigInt(value.pkg_height),
+      });
+      return Any.create({
+        type_url: MsgEndpoint.MSG_ENABLE_PKG,
```
junghoon-vans, line 350: Could we also wire these two message types through the extension transaction pipeline? `AdenaExecutor.validateMessages` and `validateInjectionTransactionType` still reject `/vm.m_enable_pkg` and `/vm.m_reject_pkg`, and `mappedDocumentMessagesWithCaller` throws `Unsupported transaction message type` for both. I confirmed the mapper rejection with both message types on this commit, even though their module encode/decode round trips succeed.

As a result, dApp requests fail before reaching these new encoders. Please update the extension message types, validators (including the `approver`/`sender` address checks), caller mapping, approval rendering, and multisig mappings alongside these codecs. If this PR intentionally adds codec-only support, please make that limitation explicit rather than implying the messages are usable through the extension.

jinoosss: Thanks, good catch. Wired both messages through the extension pipeline in the gno-js-client 2.2.0 commit (7b0e8773): inject message types, request validators with the `approver` / `sender` address checks, injection allow-lists, caller mapping (`mappedTransactionMessages` / `mappedDocumentMessagesWithCaller` / `mappedRawTxMessages`), approve popup renderers, broadcast and multisig broadcast screens, sign-multisig caller extraction and the multisig raw-tx mapper. Added tests for the validators and the caller mapping. Session accounts still cannot sign these (session-signing-guard unchanged), which matches the other admin-style messages.

junghoon-vans (approved):

PR #905, opened 2026-09-17
feat(grc20): send transfers as MsgCall when the token publishes a route

jinoosss:
## Summary

Sends a GRC20 transfer as a direct `MsgCall` into the token's own realm when the token publishes a `routes` entry, falling back to the existing `MsgRun` path otherwise.

`grc20reg` cannot move a user's balance by design: its write wrapper acts as the *calling* realm, and `grc20.CallerTeller` — the only teller that acts for whoever crossed in — sits behind the home-realm guard on `*PrivateLedger`. That is the confused-deputy defense and it is not going away, so a user-facing write has to go to the token's own realm.

The destination realm is derivable from `pkg_path`. The **call shape** is not, and it is not uniform across realms. [onbloc/gno-token-resource#73](https://github.com/onbloc/gno-token-resource/pull/73) adds an optional `routes` field that declares it:

```json
"routes": {
  "funcs": {
    "transfer":      { "name": "Transfer",     "args": ["$to", "$amount"] },
    "approve":       { "name": "Approve",      "args": ["$spender", "$amount"] },
    "transfer_from": { "name": "TransferFrom", "args": ["$from", "$to", "$amount"] }
  }
}
```

The message is assembled as `MsgCall{ PkgPath: pkgPath, Func: funcs[op].name, Args: funcs[op].args.map(substitute) }`.

## Transfer message priority

1. **`routes` MsgCall** — direct into the token realm. Does not depend on a helper realm existing on the chain.
2. **`helperPath` MsgCall** — the chain's GRC20 helper realm, unchanged.
3. **`grc20reg` MsgRun** — unchanged fallback.

`routes` is not yet published, so behavior is identical until onbloc/gno-token-resource#73 lands.

## Design notes

**`args` is the full positional list, not a prefix.** It carries every argument except the leading `cur realm`, in declaration order. A `$`-prefixed element is a placeholder (`$to`, `$from`, `$spender`, `$owner`, `$amount`); anything else is a literal passed through verbatim, with `$$` escaping a literal `$`.

Carrying the whole list matters for factory realms, where one realm hosts several tokens and takes a symbol argument. Symbol-first is a convention — neither `p/nt/grc20` nor any contract enforces it — so a literal may sit anywhere:

```json
"transfer": { "name": "Transfer",    "args": ["FOO", "$to", "$amount"] }
"transfer": { "name": "Transfer",    "args": ["$to", "$amount", "FOO"] }
"transfer": { "name": "VoucherSend", "args": ["$to", "ibc/27394FB0.../FOO", "$amount"] }
```

**Routes are not threaded through the stored token model.** Token metainfos are persisted per account and can predate this field, so reading routes from storage would make how a transfer is signed depend on when the token was first saved. `useGRC20Routes` reads the resource file directly (react-query, per network) and the transfer message consults it at build time. No storage migration.

**A route that cannot be satisfied falls back rather than signing.** An unknown placeholder (a typo such as `$reciever`) or a placeholder with no value returns `null`, and the helper/MsgRun paths still apply. Treating an unknown `$name` as a literal would silently put a wrong argument on-chain.

**Only `transfer` is wired.** The wallet does not currently build `Approve` or `TransferFrom` messages. The resolver is operation-agnostic, so adding one is wiring only.

## Forward compatibility with `grc20routes`

[gnolang/gno#6187](https://github.com/gnolang/gno/pull/6187) (`r/nt/grc20routes/v0`) aims to solve this on-chain, but is not merged and not deployed on `gnoland-1` (verified via `vm/qfile`). An on-chain route converts into this shape mechanically as `args = [...prefix, ...opTail(op)]`. When on-chain lookup lands, the resolver priority becomes `on-chain declared/proven → resource JSON → MsgRun`, and only `useGRC20Routes` changes.

## Removed: IBC token resource plumbing

`gno-token-resource` has only `gno-native/` and `grc20/` — the `ibc-native/` and `ibc-tokens/` directories no longer exist, and the fetches were already commented out.

Removed: `fetchIBCNativeTokenAssets` / `fetchIBCTokenAssets`, `TokenMapper.fromIBCNativeMetainfos` / `fromIBCTokenMetainfos`, `IBCNativeTokenResponse` / `IBCTokenResponse`, `IBCNativeTokenModel` / `IBCTokenModel`, `isIBCNativeTokenModel` / `isIBCTokenModel`, and `'ibc-native' | 'ibc-tokens'` from `TokenModel['type']`.

Deliberately untouched:

- **The IBC transfer UI mode** (`transfer-mode-tabs`, `transfer-input`, the ADN-760 TODOs) — a separate in-progress feature, unrelated to the token resource.
- **`migrates/migrations/v002`–`v022`** — frozen storage snapshots; editing their types would change what a migration means.

## Testing

- `src/common/utils/grc20-route.spec.ts` — 14 cases: placeholder substitution, leading literal (factory), trailing literal, `$$` escape, unknown-placeholder rejection, missing value, `transfer_from` shape, empty `funcs`, absent `routes`, and per-operation drop of a malformed route.
- Full suite: **955 passed**, up from 941 on `main` (+14 new).
- 17 suites / 32 tests fail identically on `main` and on this branch — migration specs hitting libsodium `invalid salt length`, an environment issue unrelated to this change. Verified by running the suite in a clean worktree at `HEAD`.
- `tsc --noEmit`: no new errors. The single remaining error (`adena-module/src/ledger/ledger-app-utils.ts`) is pre-existing.
- ESLint: no errors on changed files.


packages/adena-extension/src/pages/popup/wallet/transfer-summary/index.tsx
```diff
@@ -63,16 +70,13 @@ const TransferSummaryContainer: React.FC = () => {
...
-    cosmosProvider,
-    sessionRepository,
-  } = useAdenaContext();
+  const { transactionService, chainRegistry, tokenRegistry, cosmosProvider, sessionRepository } =
+    useAdenaContext();
   const queryClient = useQueryClient();
   const { currentAccount, currentAddress, currentFundingAddress } = useCurrentAccount();
   const { currentNetwork } = useNetwork();
+  // Per-token MsgCall shapes from gno-token-resource.
+  const { data: grc20Routes } = useGRC20Routes();
```
junghoon-vans: `grc20Routes` is initially `undefined` while this query is loading, so the pre-warmed document is built and simulated with the helper/MsgRun fallback. When the route data arrives, `getGRC20TransferMessage` changes, but the document-building effect does not depend on the route query state and its existing-document branch only updates the fee fields. The submit path later calls `createDocument()` again and can therefore sign the direct token-realm `MsgCall` using gas/error state produced for a different fallback message; on a chain where that fallback cannot execute, the simulation error also remains even after the valid route has loaded.

Please wait for the GRC20 route query to settle before creating the initial document, and rebuild the simulation document when the current network's route result changes. A settled query with no published route can still use the existing fallback.

jinoosss: The process has been modified to generate the signature document after the loading of `gno-token-resource` is complete.

jinoosss: [5004ead](https://github.com/onbloc/adena-wallet/pull/905/commits/5004eadbdb18337e340b59ee441af909e1dab599)

junghoon-vans (changes requested):
The fee-simulation race is fixed by `5004ead`, and I resolved that thread. One correctness issue still blocks approval:

`getGrc20RouteFunc()` returns `null` both when no route is published and when a published route deliberately omits `transfer`. `getGRC20TransferMessage()` then falls through to the helper/MsgRun path in both cases. The companion route contract distinguishes these states: no `routes` means fallback, while a missing `funcs.transfer` means the realm does not expose user transfers and the wallet must disable the action. PR #73 already publishes xGNS as `routes: { funcs: {} }`, so the current code will still construct a fallback transaction for that explicitly non-transferable token.

Please preserve this tri-state and prevent document/transaction creation when a published route has no `transfer` operation. After that is addressed, I do not see another blocker.

junghoon-vans (approved):
The route-loading and fee-simulation mismatch is fixed in `5004ead`. Missing transfer route metadata intentionally continues through the existing helper/MsgRun fallback. No remaining blockers from my review.

PR #907, opened 2026-09-17
feat(rpc): fail over to an optional fallbackRPCUrl when the primary RPC is down

jinoosss:
## Summary

Adds an optional `fallbackRPCUrl` to `chains.json`, set on `gnoland-1` to `https://rpc.gno.land:443`. When a request against the primary `rpcUrl` fails at the transport level, the provider retries on the fallback and keeps using it; if the fallback also fails, it rotates back to `rpcUrl`.

```json
{
  "id": "gnoland-1",
  "rpcUrl": "https://rpc.onbloc.xyz:443",
  "fallbackRPCUrl": "https://rpc.gno.land:443",
  ...
}
```

## Failover behaviour

`RpcEndpointSelector` (new, `common/provider/gno/rpc-endpoint-selector.ts`) treats a network's endpoints as a ring:

1. Requests start on `rpcUrl`.
2. A transport failure rotates to `fallbackRPCUrl` and retries the same request there. The rotation is **sticky**, so the retry and every later request go straight to the fallback instead of paying the dead endpoint's timeout again.
3. A transport failure on `fallbackRPCUrl` rotates back to `rpcUrl` in the same way.
4. Each `run` tries every endpoint at most once, so a network that is down everywhere still fails fast rather than looping.

Only **transport** failures rotate — an unreachable host, a 5xx/429 response, a timeout. Errors the chain actually answered with (insufficient funds, a rejected tx, a 4xx on a malformed query) are rethrown untouched: rotating on those would move off the primary endpoint over a failure the fallback is going to reproduce anyway.

## Coverage

`GnoProvider` issues requests through several transports, and all of them now go through the selector:

- its own `fetch` / `axios` / `RestService` calls (`getStatus`, `getGasPrice`, `getSessions`, `getSession`, `sendTransactionSync`, `simulateTx`, `getRealmDocument`);
- the base `GnoJSONRPCProvider`'s `Tm2Client`, which is now constructed with a `FallbackRpcClient` (an `RpcClient` implementation wrapping one `HttpClient` per endpoint) so inherited methods such as `getAccount` and `evaluateExpression` fail over too.

`fetchABCIResponse` / `postABCIResponse` now raise on an HTTP status >= 400, reusing tm2-rpc's `Bad status on response: <code>` wording. Previously a 502 gateway page surfaced as an opaque JSON parse error, which the selector could not classify as a dead endpoint.

Provider construction sites updated to pass the field: `AdenaProvider`, `WalletProvider`, `InjectCore.initGnoProvider`, and `WalletAccountService.getAccountInfoByNetwork`.

## No migration

`ChainRepository.getNetworks()` prefers the locally stored entry when it shares an id with a bundled default, so an existing install's `gnoland-1` entry would never see the new field. Instead of a migration, `getNetworks` grafts `fallbackRPCUrl` from the bundled `chains.json` onto stored default entries — but **only while the stored `rpcUrl` still matches the bundled one**. A user who repointed a default network at their own node keeps no fallback and is never silently failed over to ours.

## Health check

`fetchHealth` takes an optional `fallbackUrl` and only reports the network as unresponsive once neither endpoint answers, matching the provider's failover. Wired into the "network unresponsive" banner in `use-network`.

The one-off pre-flight health checks in the approve/broadcast popups are left alone for now: they also run against dApp-supplied custom networks, which have no fallback to thread through.

## Tests

- `rpc-endpoint-selector.spec.ts` (new) — rotation, stickiness, rotating back after the fallback fails, 5xx vs 4xx classification, chain-level errors not rotating, no-fallback and duplicate-fallback cases.
- `gno-provider.spec.ts` — `getSessions` retries on the fallback endpoint when the primary is unreachable.
- `chain.spec.ts` — `getNetworks` grafts the bundled fallback onto a pre-existing entry, and drops it once the user repoints the network.

`tsc --noEmit` and `eslint` report no new issues; the 62 suites / 480 tests under `common`, `repositories`, `inject`, `hooks` and `services` pass.

> Note: the `migrates/v002`–`v018` specs fail with libsodium `invalid salt length`. Verified to fail identically on a clean checkout of `main` — pre-existing and unrelated to this change.


jinoosss:
## Security review

Reviewed the failover paths for endpoint injection, traffic redirection and transaction-safety issues. One real issue found and fixed in `271a0b3`; the rest checked out.

### Fixed — stale fallback survived a network repoint (privacy)

`edit-custom-network` built the updated network by spreading the origin entry, so editing `gnoland-1`'s RPC to a private/local node kept `fallbackRPCUrl: https://rpc.gno.land:443` on the object. `getNetworks` sanitises that on the next read, but `updateNetwork` hands the in-memory object straight to `changeNetworkOfProvider`, so until that read the wallet would have failed over from the user's own node to a public endpoint they explicitly opted out of — leaking addresses, queries and broadcasts.

The edit now drops the fallback whenever `rpcUrl` changes, matching the rule the read path already enforced.

### Checked — no issue

**A dApp cannot inject a `fallbackRPCUrl`.** `inject/message/methods/network.ts` reads only `chainId` / `chainName` / `rpcUrl` from the message, and the approval popup calls `addNetwork('gno', chainName, rpcUrl, chainId, …)` → `ChainService.addGnoNetwork`, which builds a fresh object literal. No request data is spread into a `NetworkMetainfo`, so the field can only originate from the bundled `chains.json`.

**Custom networks never inherit the fallback.** `getNetworks` only grafts onto entries whose id matches a bundled default; custom entries pass through the filter untouched.

**Rebroadcasting a transaction on the fallback is safe.** The retry sends the same pre-signed bytes, and the signature covers `account_number` + `sequence` + `chain_id`, so a second delivery is rejected as a duplicate rather than executed twice. No double-spend path.

**No silent HTTPS downgrade.** `fallbackRPCUrl` carries an explicit `https://`. Had it not, tm2-rpc's `HttpClient` constructor throws `Endpoint URL is missing a protocol`, which `isTransportError` classifies as non-transport and rethrows — it surfaces as an error rather than degrading to plaintext.

**Rotation cannot be weaponised.** A rotation only ever moves between the two endpoints already configured for the network, so a node that forces one (e.g. by returning a 5xx) gains nothing — it can push the wallet onto the other configured endpoint, not onto one of its choosing.

**No ReDoS.** `TRANSPORT_MESSAGE_PATTERN` is a flat literal alternation with no nested quantifiers, matched against error messages only.

### Noted, not a risk

- The fallback endpoint sees the same request content as the primary — addresses, queries, signed transactions. `rpc.gno.land` is the official gno.land RPC, so this is a second first-party observer of activity the primary already sees, not a new exposure class. Worth being explicit about since it is a behavioural change for `gnoland-1` users.
- `toGnoNetworkProfile` still registers only `rpcEndpoints: [rpcUrl]` in `chainRegistry`, so consumers going through the registry rather than `GnoProvider` do not fail over. A coverage gap, not a security issue.

jinoosss:
## Correction to the security review above

The "stale fallback survived a network repoint" fix in `271a0b3` has been reverted in `95a01b4`, because the rule it enforced also broke the feature itself.

`getNetworks` only grafted `fallbackRPCUrl` onto a stored default entry while its `rpcUrl` still matched `chains.json`. Editing `gnoland-1`'s RPC from the network screen — the natural way to exercise the failover — therefore dropped the fallback, and no request ever reached `rpc.gno.land`.

`fallbackRPCUrl` is bundled config keyed by network id, so it is now taken from `chains.json` unconditionally for default networks. Custom networks still never receive one.

**Tradeoff, stated explicitly:** editing a default network's RPC URL no longer removes its bundled fallback. If a user repoints `gnoland-1` at a private or local node and that node is unreachable, the wallet fails over to `rpc.gno.land:443` — an endpoint the edit screen does not show. The chain is the same and the endpoint is the official gno.land RPC, so this is not a new trust boundary, but it is a behaviour worth surfacing in the UI later (e.g. showing the fallback on the edit screen, or an explicit per-network toggle).

Every other finding in the review above is unaffected: a dApp still cannot inject a `fallbackRPCUrl`, custom networks never inherit one, rebroadcasts remain sequence-bound, and there is no HTTPS downgrade path.

jinoosss:
## Disregard the previous comment — the guard stays

`95a01b4` (which removed the rpcUrl-match guard) has been reverted in `42f71cc`. The behaviour described in the security review two comments up is the behaviour on this branch:

`fallbackRPCUrl` applies to a default network only while its stored `rpcUrl` still matches the bundled `chains.json`. Editing a default network's RPC URL from the network screen drops the fallback, so a user who repoints `gnoland-1` at their own node is never failed over to `rpc.gno.land`.

### Consequence for testing

Editing the RPC URL in the wallet UI cannot exercise the failover — the edit itself removes the fallback. Use one of these instead:

- **DevTools request blocking** (Network → right-click a request → Block request domain, or the Request blocking panel with `*rpc.onbloc.xyz*`). Leaves `chains.json` and stored networks untouched, and a blocked request fails at the transport layer exactly like an unreachable host, so the retry should go to `rpc.gno.land:443`.
- **Edit `chains.json`'s `rpcUrl` to a dead host, rebuild, and start from cleared storage.** A stored network entry wins over the bundled one, so an existing profile would keep using `rpc.onbloc.xyz` and never reach the dead host.

notJoon (approved):

